PR #6, opened 2021-02-08
Improve serialization performance by avoiding the use of erased-serde.

### LINKED ISSUE #3: ShapeType precludes custom shapes
`ShapeType`'s options are all rather specific, preventing an implementer of `Shape` from introducing significantly diverse custom shapes. A `ShapeType::Custom` might be missing, or perhaps `Shape::shape_type` should return `Option`.

sebcrozet:
It turns out that our current serialization method for collider shapes (based on `erased-serde`) is twice as slow as the serialization of shapes represented as enums. Therefore, this PR adds a methods to all `Shape` in order to transform them to an enum `TypedShape` which can then be serialized more efficiently. This also simplifies the serialization/deserialization code significantly.

Now `TypedShape` is kind of a more descriptive enum than the existing `ShapeType`, so we may want to get rid of `ShapeType` in the future. Though I'm keeping that change for another PR since I want to think about a little more considering `ShapeType` can be converted to an integer which can be useful for collision dispatch( even though we don't use it currently).

While adding the `TypedShape` enum, I also fixed the `ShapeType` enum to add the `Custom` variant.
Fix #3.


sebcrozet:
Looks like this repository didn't have any CI. So I added one.

PR #9, opened 2021-02-21
implement TODO for <Triangle as Shape>::ccd_thickness

LoganDark:
This is just about the lamest change I could make to qualify as a "contributor", but I saw this TODO and thought it was something I could implement so here it is.

`cargo test` passes of course.

### DIFF
--- a/src/shape/shape.rs
+++ b/src/shape/shape.rs
@@ -543,8 +543,16 @@ impl Shape for Triangle {
     }
 
     fn ccd_thickness(&self) -> Real {
-        // TODO: in 2D use the smallest height of the triangle.
-        0.0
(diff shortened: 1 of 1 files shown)


LoganDark:
Quick note, there are a few different things that `TODO` comment could have meant, such as:
1. the minimum distance an object aimed at some part of the triangle would have to travel in order to make it through the triangle (so 0.0, naturally, because of the points)
2. the diameter of the largest circle that could fit inside the triangle, which is an ok approximation of number 1 without the points causing it to always be `0`
3. the minimum side length
4. some other estimate? maybe the minimum distance from an edge to a point?

I just went with the last one, but it's totally okay if something else was intended. Let me know and I might be able to do that instead.

PR #13, opened 2021-02-26
Add special cases for the projection of a point on a cone or a cylinder.

sebcrozet:
This makes the projection of a point on a cone or a cylinder much more accurate and faster.


sebcrozet:
The CI fails because of warnings. They will be fixed by #14.

PR #27, opened 2021-04-19
Remove debug print in halfspace contact manifold

LPGhatguy:
Hello, I just got started with rapier3d (and by extension, parry) in my project! During every step in my project, my output is flooded by what looks like a debug `println!` left in the latest parry release:

```
Found contacts: 0
Found contacts: 0
Found contacts: 0
Found contacts: 0
Found contacts: 0
Found contacts: 0
Found contacts: 0
...
```

I tracked down the print to here. Should it be removed?


Waridley:
Looks like this was covered by #29

LPGhatguy:
So it is! Alright then. I'm a little sad that there was no maintainer response!

sebcrozet:
Hey @LPGhatguy! It looks like I overlooked your PR. Really sorry for not giving any response, and thank you very much for taking the time to make that PR.

PR #32, opened 2021-05-28
Check all signs are equal

MJohnson459:
Fix a bug when tracing a ray against a triangle in 2d. I've tested this locally and it seems to fix the failure case. Existing tests should be passing.

Is there somewhere I should add new tests to cover this case?

Discussed on discord:
https://discord.com/channels/507548572338880513/748627261384556715/847755138641952800

### DIFF
--- a/build/parry2d/tests/geometry/ray_cast.rs
+++ b/build/parry2d/tests/geometry/ray_cast.rs
@@ -1,7 +1,7 @@
 use na::{self, Isometry2, Point2, Vector2};
 use parry2d::math::Real;
 use parry2d::query::{Ray, RayCast};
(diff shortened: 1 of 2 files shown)


sebcrozet:
Thank you for this PR! For clarity, the variables should be renamed to `perp_sign1`. 
You can add the new tests in the `build/parry2d/tests/geometry/ray_cast.rs` file.

MJohnson459:
Added two tests to cover the starting inside and starting outside cases. I confirmed that the tests fail with the old check and pass with the new. Also updated the variable names.

sebcrozet:
Thanks!

PR #36, opened 2021-06-10
Introduce ScaledTriMesh shape

Ralith:
Allows reuse of collision geometry between scaled objects, e.g. encoded in a GLTF file.

I think this change is more or less complete, albeit untested, but I'd like to discuss some questions before moving forwards:

- Should the transformation be more general? Translation/rotation can be expressed elsewhere, but maybe someone wants to skew? GLTF at least doesn't support anything beyond translation + rotation + nonuniform scaling, but there's not much downside to an arbitrary transform matrix here.
- Should the transformation be *less* general? Uniform scaling could in theory allow sharing of the quadtree, as well as the actual geometry, although this would require additional refactoring to realize, likely involving changes to `QBVH` and perhaps its existing users. Losing support for some cases of GLTF is a shame, but maybe nobody uses nonuniform scaling much in practice?
- Is any of this worth the extra code? Downstream code can always just make another `TriMesh` with whatever vertices it likes. Collision geometry usually isn't that big, especially for repeated objects. The composite shape abstraction makes the implementation pretty concise, but even so...

### DIFF
--- a/src/bounding_volume/bounding_sphere_trimesh.rs
+++ b/src/bounding_volume/bounding_sphere_trimesh.rs
@@ -1,8 +1,22 @@
 use crate::bounding_volume::BoundingSphere;
 use crate::math::{Isometry, Real};
-use crate::shape::TriMesh;
+use crate::shape::{ScaledTriMesh, TriMesh};
 
 impl TriMesh {
+    /// Computes the world-space bounding sphere of this triangle mesh, transformed by `pos`.
+    #[inline]
+    pub fn bounding_sphere(&self, pos: &Isometry<Real>) -> BoundingSphere {
+        self.local_aabb().bounding_sphere().transform_by(pos)
+    }
+
+    /// Computes the local-space bounding sphere of this triangle mesh.
+    #[inline]
+    pub fn local_bounding_sphere(&self) -> BoundingSphere {
+        self.local_aabb().bounding_sphere()
+    }
+}
+
+impl ScaledTriMesh {
     /// Computes the world-space bounding sphere of this triangle mesh, transformed by `pos`.
     #[inline]
     pub fn bounding_sphere(&self, pos: &Isometry<Real>) -> BoundingSphere {
--- a/src/query/point/point_composite_shape.rs
+++ b/src/query/point/point_composite_shape.rs
@@ -7,8 +7,8 @@ use crate::query::{
     visitors::CompositePointContainmentTest, PointProjection, PointQuery, PointQueryWithLocation,
 };
 use crate::shape::{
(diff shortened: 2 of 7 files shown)


MOZGIII:
Can we merge this?

Ralith:
It should be tested first, and I'd like to see the discussion points addressed.

Jasper-Bekkers:
> Should the transformation be more general? Translation/rotation can be expressed elsewhere, but maybe someone wants to skew? GLTF at least doesn't support anything beyond translation + rotation + nonuniform scaling, but there's not much downside to an arbitrary transform matrix here.

My usecase is exactly this, our data comes from regular gltf files and so we inherit their transforms and need to pass those onto runtime systems like the physics system. Switching to uniforrm scaling wouldn't help us.


> Is any of this worth the extra code? Downstream code can always just make another TriMesh with whatever vertices it likes. Collision geometry usually isn't that big, especially for repeated objects. The composite shape abstraction makes the implementation pretty concise, but even so...

We're actually running into this - having to runtime duplicate triangle data just to get a scaled clone of the TriMesh, having this upstreamed / merged would benefit us for sure.

Ralith:
> We're actually running into this - having to runtime duplicate triangle data just to get a scaled clone of the TriMesh

The question is, is doing that duplication causing you any concrete problems?

Jasper-Bekkers:
> > We're actually running into this - having to runtime duplicate triangle data just to get a scaled clone of the TriMesh
> 
> The question is, is doing that duplication causing you any concrete problems?

Yes. We have transforms changing at runtime, including scale. It's both CPU and memory overhead we want to avoid.

Jasper-Bekkers:
Any update on this?

bourumir-wyngs:
URDF (unified robot description format) has per-axis scaling attribute for meshes. It is important as some meshes come in meters, others in millimeters, have even seen one in inches.

Ralith:
Those don't seem like per-axis properties, and also sound like something you'd be best off handling *before* constructing the `TriMesh`.

bourumir-wyngs:
URDF defines the initial transform of the mesh in the form like

<visual>
  <origin xyz="0.1 0.2 0.3" rpy="0.0 1.5708 0.0" />
  <geometry>
    <mesh filename="package://path_to_your_mesh/mesh_file.ply" scale="1.0 1.0 1.0" />
  </geometry>
</visual>

This must be done when the mesh is first loaded. The ply-rs reader we use is more like a parser and does not offer these options.

Ralith:
I'm not sure what options you have in mind. You should parse that transform, then apply it to points before using them to construct a parry `TriMesh`.

PR #39, opened 2021-06-21
Clarify semantics of TOI members in Penetrating status

Ralith:
This caught me by surprise. For future work, we might consider adjusting the TOI query interface such that the `Penetrating` case doesn't return data at all.


sebcrozet:
Thanks!

PR #41, opened 2021-07-08
Fix QBVH crash when nodes are empty

skairunner:
QBVH traversal panics when `nodes` is empty. I add a check to prevent a panic.

### DIFF
--- a/src/partitioning/qbvh.rs
+++ b/src/partitioning/qbvh.rs
@@ -415,6 +415,10 @@ impl<T: IndexedData> QBVH<T> {
         stack.clear();
         stack.push(0);
 
+        if self.nodes.is_empty() {
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


src/partitioning/qbvh.rs
```diff
@@ -415,6 +415,10 @@ impl<T: IndexedData> QBVH<T> {
         stack.clear();
         stack.push(0);
 
+        if self.nodes.is_empty() {
+            return;
+        }
+
```
sebcrozet, line 421: Let's not do a useless push if there is no node.

```suggestion
        stack.clear();

        if !self.nodes.is_empty() {
            stack.push(0);
        }
```

sebcrozet (changes requested):
Thank you for this PR!

sebcrozet:
Thanks!

PR #43, opened 2021-07-24
Add RayCast impl to SharedShape

FredrikNoren:


src/query/ray/ray_shared_shape.rs
```diff
@@ -0,0 +1,103 @@
...
+          }
+        }
+        #[cfg(feature = "dim2")] {
+          if let Some(shape) = self.as_round_convex_polygon() {
+            shape.cast_local_ray(ray, max_toi, solid)
+          } else {
+            panic!("Unsupported shape");
+          }
+        }
+      }
```
Demindiro, line 50: Why not just `self.0.as_ref().cast_local_ray(ray, max_toi, solid)`?

Also, you should run `cargo fmt` This project uses 4 space indentation, not 2.

sebcrozet:
Thank you for this PR!
Is there a particular reason for adding this trait implementation? The `Shape` trait already derives from `RayCast`, which means that `&*shared_shape.0` implements `RayCast`.

src/query/ray/ray_shared_shape.rs
FredrikNoren: @Demindiro Ah true, much better. Updated

FredrikNoren:
@sebcrozet Just for consistency; I'm using `Box<dyn RayCast>` in my project and I wanted to be able to dump anything in there that can be ray cast (shared or "simple").

Demindiro:
> The `Shape` trait already derives from `RayCast`, which means that `&*shared_shape.0` implements `RayCast`.

> I'm using `Box<dyn RayCast>` in my project and I wanted to be able to dump anything in there that can be ray cast (shared or "simple").

Perhaps it would be useful to to implement `Deref` with `Target = &dyn Shape`? That way all methods implemented on `Shape` should be accessible without needing `&*shared_shape.0`.

FredrikNoren:
@Demindiro Looks like `type Target = dyn Shape;` is already implemented on SharedShape. Not sure if `type Target = &dyn Shape;` is different though?

Demindiro:
> Looks like `type Target = dyn Shape;` is already implemented on SharedShape. Not sure if `type Target = &dyn Shape;` is different though?

Indeed. I would expect `cast_local_ray` to be already useable then though.

FredrikNoren:
@Demindiro Hm, I tried just doing `Box::new(shared_shape.clone())` but I got `the trait bound 'SharedShape: RayCast' is not satisfied. required for the cast to the object type `dyn RayCast`. Not sure if there's some other way to do that though. `Box::new(pick_bounding.clone() as dyn RayCast)` doesn't work either.

PR #45, opened 2021-08-01
fix benchmark cargo config and update API calls to match most recent …

dbenson24:
…parry version


sebcrozet:
Thank you. Sorry for getting to this late. This is no longer relevant.

PR #57, opened 2021-10-18
Add bevy-components feature

deontologician:
After https://github.com/bevyengine/bevy/pull/2254 there is a requirement that anything used as a component must derive `Component` explicitly.

From what I can tell, most of the data structures that are used as components in `bevy_rapier` are defined in `rapier`, but `SharedShape` is re-exported from `parry`, so the annotation has to go here.

I attempted to minimize the effect on `parry` by:
- putting it behind a feature flag
- minimizing the number of lines of code referring to bevy in parry
- turning off default features in bevy to reduce the effect on compile time

One thing I'm unsure of is specifying the version of bevy. I've included it as `0.5` here, because that's what's on crates.io right now, but this code won't actually work with that version, only with a particular branch of bevy in this PR (which will hopefully land soon):

https://github.com/bevyengine/bevy/pull/2305

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,3 +6,4 @@ parry2d = { path = "build/parry2d" }
 parry3d = { path = "build/parry3d" }
 parry2d-f64 = { path = "build/parry2d-f64" }
 parry3d-f64 = { path = "build/parry3d-f64" }
+bevy = {path = "../bevy"}
--- a/build/parry2d-f64/Cargo.toml
+++ b/build/parry2d-f64/Cargo.toml
@@ -27,6 +27,7 @@ enhanced-determinism = [ "simba/libm_force", "indexmap" ]
 # Do not enable this feature directly. It is automatically
 # enabled with the "simd-stable" or "simd-nightly" feature.
 simd-is-enabled = [ ]
+bevy-components = ["bevy"]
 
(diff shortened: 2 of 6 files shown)


MiniaczQ:
I don't think a feature like this should belong in the main branch. It's only a temporary fix until we get better syntax to do it properly from only the interop crate.

Aceeri:
@MiniaczQ I'm not sure there is much of an alternative to this currently (maybe doing just wrapper types in the interop crate?). I don't think we are getting better syntax for doing this anytime soon based on https://github.com/rust-lang/rfcs/pull/2393.

deontologician:
> @MiniaczQ I'm not sure there is much of an alternative to this currently (maybe doing just wrapper types in the interop crate?). I don't think we are getting better syntax for doing this anytime soon based on looking at https://github.com/rust-lang/rfcs/pull/2393.

It sounds like wrappers in the interop crate is the way to go, but that requires the most knowledge of the innards. This change was relatively easy to make as an outsider without changing how `bevy_rapier` works too internally.

sebcrozet:
Thank you for this PR @deontologician! I’m closing it since I ended up using wrapper types instead: https://github.com/dimforge/bevy_rapier/pull/102.

PR #60, opened 2021-10-23
Create `TriMesh` from (non-convex) polygon

### LINKED ISSUE #59: Create trimesh from non-convex 2D polygons
I'd like to be able to create a `TriMesh` from non-convex 2D polygons.

I have a PR ready implementing the ear-clipping algorithm. Once we have this, we can think about adding the Hertel-Mehlhorn algorithm that combines the triangles into convex polygons again, to give a good convex decomposition.

The reason for this is that, at least in my experiments, the VHACD algorithm is orders of magnitude slower than this method.

EmbersArc:
Closes #59.

I didn't quite know where best to put things so feel free to make suggestions. The algorithm is largely based on https://github.com/ivanfratric/polypartition.


sebcrozet (approved):

sebcrozet:
Perfect, thanks!

PR #61, opened 2021-10-28
Remove unneeded comparison for 2D intersections test

GuillaumeGomez:
At this point I think it's nitpicking but I still think it could be useful.


sebcrozet:
Thanks!

PR #63, opened 2021-10-31
Add Hertel-Mehlhorn algorithm

### LINKED ISSUE #62: Create compound shape from 2D trimesh (Hertel-Mehlhorn algorithm)
#59 added non-convex polygon to `TriMesh` conversions. For more efficient operations, we want to turn those `TriMesh`es into `Compound` shapes, while merging neighboring triangles into convex polygons. I have a pull request ready that achieves this using the Hertel-Mehlhorn algorithm.

EmbersArc:
Closes #62.

Again, I'm not sure how to make this accessible to the user. I've added it as a method to `TriMesh` for now, but maybe adding it as `CompoundShape::from_trimesh()` makes more sense.

### DIFF
--- a/src/shape/compound.rs
+++ b/src/shape/compound.rs
@@ -5,7 +5,11 @@
 use crate::bounding_volume::{BoundingSphere, BoundingVolume, AABB};
 use crate::math::{Isometry, Real};
 use crate::partitioning::QBVH;
+#[cfg(feature = "dim2")]
+use crate::shape::{ConvexPolygon, TriMesh, Triangle};
 use crate::shape::{Shape, SharedShape, SimdCompositeShape, TypedSimdCompositeShape};
+#[cfg(feature = "dim2")]
+use crate::transformation::hertel_mehlhorn;
 
 /// A compound shape with an aabb bounding volume.
 ///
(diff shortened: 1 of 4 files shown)


sebcrozet (changes requested):
Thank you for this PR, this is fantastic!

I suggest the following, to keep it coherent with some other parts of the crate (like `convex_hull2`):
- Rename the `hertel_mehlhorn` function to `hertel_mehlhorn_idx`.
- Add a new `hertel_mehlhorn` function to the `hertel_mehlhorn.rs` file. This new function returns a `Vec<Vec<Point<Real>>` (it just calls `hertel_mehlhorn_idx` and replaces all the indices by the corresponding points).
- Re-export both `hertel_mehlhorn` and `hertel_mehlhorn_idx` from the `transformation_module`.
- Like you suggested, I think it is better to add a `CompoundShape::from_trimesh()` instead of adding the method to `TriMesh`.

EmbersArc:
Thanks for the review @sebcrozet! I've applied the suggestions in 7d9026b.

sebcrozet (approved):

sebcrozet:
Perfect, thank you!

PR #64, opened 2021-11-06
Treating initial mesh calculation as fallible

bellwether-softworks:
I'm attempting to naïvely perform convex hull transformations to convert point clouds to meshes, but frequently encounter a shape that results in a panic.  It would be helpful to know in advance that a shape isn't regarded as valid from Parry's perspective, so to support this I'm handling potential points of failure with an error struct.  The original `get_initial_mesh` function now calls this modified method with an `.unwrap()` call, thus preserving the original function contract.

:bulb: Ideally, the convex hull transformation would similarly allow error handling, but this serves as an interim solution.

I've also added a test for a point cloud known to result in failure in an `f32` environment.

**Edit:** I've also exposed some internals to allow for external consumers to take advantage of this behavior (in particular, as can be seen in the test, a consumer needs access to `normalize` as well as `try_get_initial_mesh`).  My approach is admittedly heavy-handed, and I can revisit this as necessary.


sebcrozet:
Hey thank you for this PR! It took me forever to get to it, but it can be merged now. I changed it to make the 3D convex-hull calculation itself faillible (with `try_convex_hull`) instead of exposing internals to the user.

PR #80, opened 2022-05-20
Add support for f64 to simd-nightly

royalmustard:
This adds support for using simd with the f64 versions on nightly.


sebcrozet:
Thank you. Sorry for getting to this late. This is now supported through `wide`.

PR #86, opened 2022-07-16
Fix typo in transformation/utils.rs

eval-exec:
`dr` should be `down-right` and `ur` should be `up-right` in code comment


sebcrozet:
Thanks!

Issue #87, opened 2022-08-02
Non-uniform scale applied to Ball Collider ignores its own radius


Gyrobifastigium:
https://github.com/dimforge/parry/blob/c3f28389e762cf93d8a1a62337fb806450852017/src/shape/ball.rs#L40

This line should be `self.to_polyline(nsubdivs)` instead of `Self::new(0.5).to_polyline(nsubdivs)`.

PR #88, opened 2022-08-03
Use correct radius on non-uniformly scaled Ball

Gyrobifastigium:
Fixes https://github.com/dimforge/parry/issues/87.


sebcrozet:
Thanks!

PR #92, opened 2022-09-19
Default to existing normal if can't normalize

Aceeri:
Currently only doing it on `ConvexPolyhedron` colliders, but could essentially copy paste this to the rest of the scaling.


sebcrozet:
Thanks!

PR #108, opened 2022-11-20
Clarify doc comments of SegmentsIntersection::Point

wlinna:
Point-intersection only has only one intersection point, so the original doc comments "Location of the first intersection point on the first segment." and Location of the second intersection point on the second segment" are confusing. It seems like they were copied from the Segment-intersection where those descriptions make sense.


sebcrozet:
Thanks!

PR #114, opened 2023-01-03
Add functionality to intersect a `TriMesh` with a plane

plof27:
## Objective

Adds functionality to intersect a intersect a `TriMesh` with a plane, extracting any `Polyline`s that lie on the intersection. The API intentionally mimics the `TriMesh` split functions, and much of the implementation is identical.

## Changelog
- Added `TriMesh::canonical_intersection_with_plane` for intersecting with planes aligned with one of the axes
- Added `TriMesh::intersection_with_plane` for intersecting with arbitrary planes
- Added `TriMesh::intersection_with_local_plane` for intersecting with arbitrary planes in the same space as the mesh
- Added `IntersectResult` as the output type for the above functions. (Mimicing the API of the splitting functions.)
- Added `Polyline::extract_connected_components`, which splits a compound polyline into its connected components.

## Notes
- I wasn't sure if unit tests were expected, but I had already written some for my own local testing, so I figured there wasn't any reason not to include them.
- Unlike the `TriMesh` splitting functions, the triangle splitting in the new function generates an adjacency list of the final line segments. This allows us to easily construct multiple, consistently-oriented lines by traversing the list.
- The other notable difference with the impl for the splitting functions is that the `new_vertices` list starts empty, and so only contains vertices that are actually output.

### DIFF
--- a/src/query/split/split_trimesh.rs
+++ b/src/query/split/split_trimesh.rs
@@ -350,12 +350,14 @@ impl TriMesh {
     /// This will intersect the mesh by a plane with a normal with it’s `axis`-th component set to 1.
     /// The splitting plane is shifted wrt. the origin by the `bias` (i.e. it passes through the point
     /// equal to `normal * bias`).
+    ///
+    /// Note that the resultant polyline may have multiple connected components
     pub fn canonical_intersection_with_plane(
         &self,
         axis: usize,
         bias: Real,
         epsilon: Real,
-    ) -> IntersectResult<Vec<Polyline>> {
+    ) -> IntersectResult<Polyline> {
         self.intersection_with_local_plane(&Vector::ith_axis(axis), bias, epsilon)
     }
 
@@ -368,7 +370,7 @@ impl TriMesh {
         axis: &UnitVector<Real>,
         bias: Real,
         epsilon: Real,
-    ) -> IntersectResult<Vec<Polyline>> {
+    ) -> IntersectResult<Polyline> {
         let local_axis = position.inverse_transform_unit_vector(axis);
         let added_bias = -position.translation.vector.dot(&axis);
         self.intersection_with_local_plane(&local_axis, bias + added_bias, epsilon)
@@ -382,7 +384,7 @@ impl TriMesh {
         local_axis: &UnitVector<Real>,
         bias: Real,
         epsilon: Real,
-    ) -> IntersectResult<Vec<Polyline>> {
+    ) -> IntersectResult<Polyline> {
         // 1. Partition the vertices.
         let vertices = self.vertices();
         let indices = self.indices();
@@ -561,7 +563,7 @@ impl TriMesh {
         }
 
         // 3. Ensure consistent edge orientation by traversing the adjacency list
-        let mut polylines: Vec<Polyline> = Vec::new();
+        let mut polyline_indices: Vec<[u32; 2]> = Vec::with_capacity(index_adjacencies.len() + 1);
 
         let mut seen = vec![false; index_adjacencies.len()];
         for (idx, neighbors) in index_adjacencies.iter().enumerate() {
@@ -573,11 +575,9 @@ impl TriMesh {
                 let mut prev = first;
                 let mut next = neighbors.first(); // Arbitrary neighbor
 
-                let mut vertex_indices = vec![prev];
-
                 'traversal: while let Some(current) = next {
                     seen[*current] = true;
-                    vertex_indices.push(*current);
+                    polyline_indices.push([prev as u32, *current as u32]);
 
                     for neighbor in index_adjacencies[*current].iter() {
                         if *neighbor != prev && *neighbor != first {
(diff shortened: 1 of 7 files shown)


src/query/split/split_trimesh.rs
```diff
@@ -344,6 +346,274 @@ impl TriMesh {
...
+
+        // Helper function for adding polyline segments to the adjacency list
+        let mut add_segment_adjacencies = |idx_a, idx_b| {
+            let _ = index_adjacencies
+                .entry(idx_a)
+                .and_modify(|list: &mut HashSet<u32>| {
+                    let _ = list.insert(idx_b);
+                })
+                .or_insert_with(|| {
+                    let mut adjacencies = HashSet::with_capacity(2);
```
plof27, line 429: I noticed that the `HashMap`s used in other functions are re-exports of `FxHashMap`. I wasn't sure if I should use the fx variant of HashSet, so I just used the `std` version. I'm happy to change this if it's an issue.

src/query/split/split_trimesh.rs
```diff
@@ -344,6 +346,274 @@ impl TriMesh {
...
+                    let ib = (ie + 1) % 3;
+                    let ic = (ie + 2) % 3;
+                    let idx_a = idx[ia as usize];
+                    let idx_b = idx[ib as usize];
+                    let idx_c = idx[ic as usize];
+                    assert_eq!(iv, ic);
+
+                    let intersection_idx = intersect_edge(idx_a, idx_b);
+
+                    add_segment_adjacencies(idx_c, intersection_idx);
```
sebcrozet, line 542: That `idx_c` looks incorrect? It belongs to the mesh’s index space instead of indexing into the `new_vertices` array. Unless I missed something, I would expect and index conversion like in [the vertex/vertex case](https://github.com/dimforge/parry/pull/114/files#diff-69c40d9287d4877ea22122e7d1fa7509c093a5e0edb3d77571bb8116a5df2ff6R520-R525) to happen for `idx_c`.

sebcrozet (changes requested):
Thank you for this PR!

The added `TriMesh::intersection_with_local_plane` method will first calculate all the intersection points and then identify connected components of the polyline (while fixing its orientation at the same time). I suggest to split this in two methods:
1. Modify `TriMesh::intersection_with_local_plane` so it only returns a single polyline: `IntersectResult<Polyline>`.
2. Add a `Polyline::extract_connected_components` method to `Polyline` (in the `polyline.rs` file) that calculates and return the connected components with fixed orientations.

For 2, an alternative would be to define `PolylineFlags` similar to `TriMeshFlags`, but let’s keep it simple for now as long as we don’t need features like merging duplicate vertices.

src/query/split/split_trimesh.rs
plof27: I had it working like this originally, but it's possible for the same neighbor to be added twice if multiple, consecutive triangle segments lie in the intersection plane. That being said, it shouldn't actually cause any issues to have those duplicates (other than a very minor perf hit). I'll rework :)

src/query/split/split_trimesh.rs
plof27: I think you're right. Good catch! I'll add a test for this case as well.

plof27:
> The added TriMesh::intersection_with_local_plane method will first calculate all the intersection points and then identify connected components of the polyline (while fixing its orientation at the same time). I suggest to split this in two methods

Sounds good to me. I just wasn't sure if polylines with multiple components were considered valid in parry. Regardless, I'll make the change :)

src/query/split/split_trimesh.rs
plof27: Fixed as of b7693f042c3509986e10926a225cd871ad9d445b

src/query/split/split_trimesh.rs
plof27: Fixed as of 6d2cdc4b7dcffd945a86fb2b908a31e8200fe029

plof27:
`Polyline::extract_connected_components` has been added as of c48cca34a83de8a992281c8bf254dedc36381d8c. I made a few assumptions on the function because it made the impl much easier (and faster). It would certainly be possible to write a more general version, but I think that would make more sense with a `PolylineFlags` system, since the end user could specify if the assumptions are already true.

src/shape/polyline.rs
```diff
@@ -143,6 +143,55 @@ impl Polyline {
         }
     }
 
+    /// Extracts the connected components of this polyline, consuming `self`.
+    ///
+    /// Assumes that:
+    /// - This polyline is closed with `self.indices[i][1] == self.indices[(i + 1) % component_length][0]`
+    /// for each component, where `component_length` is the number of vertices in that component.
+    /// - The indices for each component are found contiguously in the index buffer.
```
plof27, line 151: This is technically a corollary of the previous statement, but I think including it improves clarity.

PR #123, opened 2023-01-24
Speed up rkyv serialization

Jasper-Bekkers:
This tries to make use of the rkyv `CopyOptimize` attribute that can guarantee memcpy-able structs are indeed just that, rather then iterating over each element in the vector.

This ran into one oddity however, the archived layout of `QbvhNode` is smaller then the regular layout (128 vs 124 bytes), this is because `QbvhNode` contains quite a few padding bytes.

 - `QbvhNodeFlags` is `u8`
 - `NodeIndex` has a `u8` member

Both of those cause a (normally) 121-bytes struct to get blown up to 128 bytes.

In this PR I've hackfixed/worked around this issue by making QbvhNodeFlags u64 which obviously isn't what anybody would want, however since I'm not familiar with this project I'd like to open the discussion about what to do with this.

It seems likely that `QbvhNode` needs to be somehow cache-line aligned (which it is at the moment), but it might be nice to mark the padding bytes in `NodeIndex` and `QbvhNode` explicitly so it's clear where in the node one can stuff some additional info.

Removing the `CopyOptimize` is not the end of the world; but in this case having it speeds up rkyv serialization by about 2x.

### DIFF
--- a/src/partitioning/qbvh/qbvh.rs
+++ b/src/partitioning/qbvh/qbvh.rs
@@ -98,7 +98,7 @@ bitflags! {
     #[cfg_attr(feature = "cuda", derive(cust_core::DeviceCopy))]
     #[derive(Default)]
     /// The status of a QBVH node.
-    pub struct QbvhNodeFlags: u8 {
+    pub struct QbvhNodeFlags: u64 {
         /// If this bit is set, the node is a leaf.
         const LEAF = 0b0001;
         /// If this bit is set, this node was recently changed.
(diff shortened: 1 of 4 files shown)


src/partitioning/qbvh/qbvh.rs
```diff
@@ -98,7 +98,7 @@ bitflags! {
     #[cfg_attr(feature = "cuda", derive(cust_core::DeviceCopy))]
     #[derive(Default)]
     /// The status of a QBVH node.
-    pub struct QbvhNodeFlags: u8 {
+    pub struct QbvhNodeFlags: u64 {
```
Jasper-Bekkers, line 101: This particular change is the one I wanted to highlight / discuss.

sebcrozet:
Thank you. Sorry for getting to this late. This is no longer relevant since QBVH has been removed.

PR #130, opened 2023-02-28
add rkyv-safe-deser feature for deriving CheckBytes on Archive types

### LINKED ISSUE #128: Add `bytecheck` validation to `Archive` types
In particular I need `Aabb: CheckBytes<C>`.

Unfortunately I think we will want to wait until a new release of `rkyv` that contains this PR: https://github.com/rkyv/bytecheck/pull/31

We *could* work around that issue by providing a manual `impl`.

samlich:
Fixes #128: Derives CheckBytes on archived types generated by rkyv, so that rkyv's safe deserialization can be used.

Adds feature `rkyv-safe-deser`, as that's how it's named in Nalgebra. 

The [Nalgebra feature](https://github.com/dimforge/nalgebra/blob/dev/Cargo.toml#L71) is marked as for tests, and it seems its regular `rkyv-serialize` feature already derives CheckBytes, so I didn't add that as a dependency.

SIMD will work, once a patch is added to Simba, but first one needs to be added to Packed-simd.


sebcrozet:
Thank you for this PR! Since `CheckBytes` impls are always enabled in `nalgebra` when `rkyv` is, I changes this PR do the same here instead of adding a new feature. That `rkyv-safe-deser` is a leftover from some debugging and should be removed now.

PR #137, opened 2023-04-04
Fixed polyline flat indices method

kurtkuehnert:
Currently, the polyline flat_indices method seems to be copied over from the Triangle shape without proper adjustment.
This PR should fix the invalid casting.


sebcrozet:
Thanks!

PR #148, opened 2023-07-30
Fix ball-convex manifolds missing contacts in corner cases

Ralith:
When a ball's origin lay directly on a surface, no contact was found. While the contact normal is ambiguous in this case, we can make up something plausible.

This was causing me considerable confusion in some unit tests, where trivially intersecting ball-triangle tests were failing with contact manifolds.


sebcrozet:
Thanks!

PR #151, opened 2023-08-05
deps: Update `num-derive` from `0.3` to `0.4`.

waywardmonkeys:
This removes the usage of `syn` v1 as it updated to using v2.


sebcrozet:
Thanks!

PR #158, opened 2023-08-27
deps: Update `indexmap` from `1` to `2`

waywardmonkeys:
This is only used when `enhanced-determinism` is enabled.

### DIFF
--- a/crates/parry2d-f64/Cargo.toml
+++ b/crates/parry2d-f64/Cargo.toml
@@ -54,7 +54,7 @@ approx          = { version = "0.5", default-features = false }
 serde           = { version = "1.0", optional = true, features = ["derive"] }
 rkyv            = { version = "0.7.41", optional = true }
 num-derive      = "0.4"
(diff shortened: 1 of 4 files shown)


waywardmonkeys:
There's an interesting detail here in that even before this, the tests fail for `cargo test --features enhanced-determinism,serde-serialize` ... but they fail same way with and without this update. (Should probably file a bug about that test failure...)

waywardmonkeys:
This fails for the cuda build as that uses a set version of the nightly from 2021 and `hashbrown`, a dependency of `indexmap` has an MSRV of 1.64.

It feels like this will become an increasingly common issue as many crates are ramping up their MSRV to more recent versions than the old nightly that Cuda support requires.

waywardmonkeys:
@sebcrozet I've rebased this forward and fixed the merge conflicts.

sebcrozet:
Fantastic, thanks!

PR #160, opened 2023-08-28
clippy: Fix mem_replace_with_default.

waywardmonkeys:
When replacing something with the default value, can juse use `take` instead.


sebcrozet:
Thanks!

PR #163, opened 2023-09-10
Fix voxel dim calc

robtfm:
square flat shapes cause the voxel dimensions to be miscalculated and allocate vastly more voxels than intended. see [v-hacd latest](https://github.com/kmammou/v-hacd/blob/454913fad7d4071f374e3d9aee3e6ccc98bca84d/include/VHACD.h#L5113)


sebcrozet:
Thanks! CI failure is not caused by this PR.

PR #165, opened 2023-09-17
If we have less than 3 points return empty convex hull

Aceeri:
Ran into this issue where providing a single point would panic rather than just return `None` when creating a convex hull collider.


sebcrozet:
Thank you! I modified the code so it returns an error instead of an empty convex hull if less than 3 points are provided.

Issue #157, opened 2023-08-26
Frustum-frustum intersection jitter


Davidster:
Hello! 

First of all: thank you for the incredible library 😄

I am experiencing some unexpected behaviour and was wondering if someone could help me troubleshoot it. I want to test if two frustums are intersecting but it seems to be producing inconsistent results as can be seen in the video below:

https://github.com/dimforge/parry/assets/2389735/8052bdf9-3d63-4dc9-b6f1-6a92eb9b098b

Here is how I am implementing the test:

```rust
impl FrustumDescriptor {
    pub fn frustum_intersection_test(
        &self,
        other: &FrustumDescriptor,
    ) -> Option<rapier3d::parry::query::contact::Contact> {
        rapier3d::parry::query::contact::contact(
            &rapier3d::na::Isometry::identity(),
            &self.to_convex_polyhedron(),
            &rapier3d::na::Isometry::identity(),
            &other.to_convex_polyhedron(),
            0.0,
        )
        .unwrap()
    }

    pub fn to_convex_polyhedron(&self) -> ConvexPolyhedron {
        let points: Vec<_> = self
            .to_basic_mesh()
            .vertices
            .iter()
            .map(|vertex| Point::from(vertex.position))
            .collect();
        ColliderBuilder::convex_hull(&points)
            .expect("Failed to construct convex hull for frustum")
            .build()
            .shape()
            .as_convex_polyhedron()
            .unwrap()
            .clone()
    }
}
```

I'm using the same to_basic_mesh() function to display the frustum in my debug setup as in the video, so I'm quite sure the vertices are OK. For full details on the implementation you can check out this [commit](https://github.com/Davidster/ikari/commit/09982380d42d85e5adeaedbffc1ad199aa58cd96)

Davidster:
Update: added unit test

Davidster:
After further investigation (and producing a failing case with intersection_test), I eventually tried with f64 and everything worked fine. I will most likely be switching to rapier_f64

Davidster:
Here is my second failing unit test for intersection_test:

https://github.com/Davidster/parry/blob/4edb6863635adf39f2b937da31ac4a82a12428b8/crates/parry3d/tests/geometry/convex_polyhedra_intersection.rs

Ughuuu:
Ok, was looking into this a bit more and read a little bit about EPA algorithm. It appears the EPA algorithm can go indefinitely as long as a better result is found (which is not the case here, at least for me it is not).
[...]

PR #167, opened 2023-10-04
return contact even in case the epa algorithm gets stuck.

Ughuuu:
Fixes https://github.com/dimforge/parry/issues/157

Still investigating if it doesn't break anything else.


Ughuuu:
Will come back if I see it working, still investigating. Seems to work, but am not 100% sure.

PR #168, opened 2023-10-13
Ensure convex polygon does not panic due to invalid amount of points

hakolao:
I'm constantly running into panics, due to

```
parry2d-0.13.5\src\shape\convex_polygon.rs:185:59:
index out of bounds: the len is 2 but the index is 2
```

This occurs more often than _normally_ (I assume), because I create compound shapes at runtime after deforming objects into smaller ones.

It should not be possible to create a polygon that has less than 3 points. Also, it should be ensured that no empty points is passed. That fixes another crash where `normals[0]` indexes points that don't exist.

### DIFF
--- a/src/shape/convex_polygon.rs
+++ b/src/shape/convex_polygon.rs
@@ -32,6 +32,9 @@ impl ConvexPolygon {
     /// Convexity of the input polyline is not checked.
     /// Returns `None` if all points form an almost flat line.
     pub fn from_convex_polyline(mut points: Vec<Point<Real>>) -> Option<Self> {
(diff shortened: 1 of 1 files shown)


caspark:
Calling Rapier's `ColliderBuilder::convex_decomposition_with_params(&vertices, &indices, &VHACDParameters { resolution: x, ..})` with 3-4 `vertices` all approximately `x` distance from each other seems to be a pretty reliable way to trigger the underlying issue here in Parry.

sebcrozet:
Thank you!

Issue #146, opened 2023-07-10
QBVH `rebalance` still panics at 'attempt to add with overflow'


Jondolf:
[Bevy XPBD](https://github.com/Jondolf/bevy_xpbd) is using Parry internally for spatial queries, and the `rebalance` method of `Qbvh` is causing panics with the error 'attempt to add with overflow'. #139 claims to fix this, but it doesn't seem like it, although the error is on a slightly different line.

```
thread 'TaskPool (6)' panicked at 'attempt to add with overflow', /home/nisevoid/.cargo/registry/src/index.crates.io-6f17d22bba15001f/parry3d-0.13.5/src/partitioning/qbvh/update.rs:367:54
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Encountered a panic in system `bevy_xpbd_3d::plugins::spatial_query::update_query_pipeline`!
Encountered a panic in exclusive system `bevy_xpbd_3d::plugins::setup::run_physics_schedule`!
Encountered a panic in system `bevy_app::main_schedule::Main::run_main`!
```

This is the code that calls `rebalance`, in the `update_incremental` method, similar to Rapier's query pipeline:

```rust
if refit_and_balance {
    let _ = self.qbvh.refit(0.0, &mut self.workspace, |entity_index| {
        // Construct entity ID (used as key in a hash map)
        let generation = self.entity_generations.get(entity_index).map_or(0, |i| *i);
        let entity = Entity::from_bits((generation as u64) << 32 | *entity_index as u64);
        // Compute and return AABB
        let (iso, shape) = colliders.get(&entity).unwrap();
        shape.compute_aabb(iso)
    });
    self.qbvh.rebalance(0.0, &mut self.workspace);
}
```

tesselode:
I am also having this issue.

hakolao:
Still getting this (just using rapier & passing query pipeline in `step`).

LachlanHogan:
Currently hitting the same issue intermittently (will come back to it in the future to fix for what I'm working on now if nobody else picks it up). Here is a stack trace:
```
thread 'main' panicked at ...\.cargo\registry\src\index.crates.io-6f17d22bba15001f\parry3d-0.13.5\src\partitioning\qbvh\update.rs:367:54:
attempt to add with overflow
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library\std\src\panicking.rs:595
[...]

PR #169, opened 2023-10-15
QbvhUpdateWorkspace stack depth change u8 to u16 to prevent overflow

honnisha:
QbvhUpdateWorkspace stack depth can be more than 255, so I changed u8 to u16.

After this change, I no longer received the error:
`attempt to add with overflow`
on the line:
`workspace.stack.push((child, depth + 1));`

Related error:
https://github.com/dimforge/rapier/issues/532


axelmagn:
bump on this? I think it would fix #146, and no discussion for half a year.  I just wrote a similar patch before seeing that there's already an open PR.

sebcrozet:
Unfortunately, this is not a valid fix. The problem here is that the incremental update tends to make the tree more progressively more degenerate (increasingly deep). Changing the depth type only hides the actual problem.
A fix here needs to rework the way rebalancing is implemented.

PR #170, opened 2023-11-05
Some parts of lib in no_std with alloc. Tests pass

hatmajster:
**This is ready, but not complete work, some things are still gated. I just tried to make it compile first, then for tests to pass. As it happened, here's the PR. I still have not used it in real product, and I did it blindly, so its a draft for now, though I don't expect much more work to be done here. Please verify carefully. Testing/review appreciated.**
I have no idea how to split this, organize etc. Any opinions, refactor ideas appreciated. I tried to not change as much code as possible, sometimes it doesn't make any sense, sometimes looks stupid. But essentially all commits should compile in some way using `no_std` and `alloc` features.
PR requires this: https://github.com/Stoeoef/spade/pull/92, so you can see this dependency in Cargo.toml for now. If it will be merged, I'll change dependency here.
As I wrote in the beginning I would like for this to not be full "no_std+alloc". Some parts are still gated. Its just a good start I hope, and if it gets merged I hope the rest may be done quickly. But I can of course add more if someone advise so. My main fear is that I just don't know if any of this actually works, so I would prefer to scope low.

### DIFF
--- a/crates/parry2d/Cargo.toml
+++ b/crates/parry2d/Cargo.toml
@@ -19,8 +19,8 @@ maintenance = { status = "actively-developed" }
 [features]
 default = [ "required-features", "std" ]
 required-features = [ "dim2", "f32" ]
-alloc   = [ "hashbrown", "num-traits/libm", "nalgebra/alloc" ]
-std     = [ "nalgebra/std", "slab", "rustc-hash", "simba/std", "arrayvec/std", "spade" ]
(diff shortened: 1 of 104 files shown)


sebcrozet:
Thank you. Sorry for getting to this late. This seems to already be fixed on master now as I cannot reproduce it. Please reopen if that’s not the case.

PR #179, opened 2024-02-11
add a Debug impl for SharedShape

whatf0xx:
Based on this issue in rapier: https://github.com/dimforge/rapier/issues/577

I implemented Debug for SharedShape by using the trait method to get the ShapeType, and then using that as the basis of the debug info. 

This is my first PR, I hope it's ok! I will post a corresponding one to rapier.

### DIFF
--- a/src/shape/compound.rs
+++ b/src/shape/compound.rs
@@ -18,7 +18,7 @@ use crate::utils::DefaultStorage;
 /// the main way of creating a concave shape from convex parts. Each parts can have its own
 /// delta transformation to shift or rotate it with regard to the other shapes.
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct Compound {
     shapes: Vec<(Isometry<Real>, SharedShape)>,
     qbvh: Qbvh<u32>,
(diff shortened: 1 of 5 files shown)


sebcrozet:
Hi! Thank you for this PR. This is a great start.
It sounds more useful to use `.as_typed_shape()` instead of `.shape_type()`, that way you can print the actual geometry instead of just the type. Perhaps `Debug` could be implemented for `TypedShape`.

whatf0xx:
Thanks for your feedback and sorry for the slow response, it's been a busy few weeks:)

I extended the `#[derive(Debug)]` to `TypedShape` and hence to all of its `enum` variants that didn't support it already. The only one which got a blanket/dummy implementation was `TriMesh`; I was confused because it threw an error trying to derive `Debug` for `Qbvh` here but elsewhere this wasn't an issue (for `Polyline`, for example). I would be happy enough to dig a little deeper and sort out a useful `Debug` implementation for `TriMesh`, too, but I'm not sure how useful that would immediately be, especially since the original issue was to just have a `Debug` implementation a few abstraction levels up...

whatf0xx:
Hi! I checked that I can now build with `cargo build -p parry2d` so this should pass the tests.

sebcrozet:
Thanks!

Issue #155, opened 2023-08-16
Add method to scale an AABB wrt its center


sebcrozet:
The `Aabb::scaled` function scales the AABB points wrt. the origin. It might be worth adding `Aabb::scaled_wrt_center` to scale relative to the aabb’s center.

sebcrozet:
Implemented in #182.

PR #182, opened 2024-03-12
Aabb wrt centre/issue #155

whatf0xx:
I wrote the function that would allow for an `aabb` to be scaled with respect to its centre instead of the origin ( Issue #155 ). I added a unit test as a sanity check that this was doing the right thing, but not anything comprehensive.


sebcrozet:
Thanks! The implementation was a bit more complicated than necessary, so I simplified it.

PR #187, opened 2024-03-29
Fix some docstrings in `tetrahedron.rs`

Chubercik:
When reading [docs about the `parry3d::shape::Tetrahedron` struct](https://docs.rs/parry3d/latest/parry3d/shape/struct.Tetrahedron.html) I noticed that the `Tetrahedron`'s `new()` method had the following description:

> Creates a tetrahedron from three points.

This PR fixes that oversight (and some others).


sebcrozet:
Thanks!

PR #196, opened 2024-05-01
Treat TriMesh as solid in project_local_point_and_get_feature with dim2

wlinna:
Fixes the issue in 2D where `project_local_point_and_get_feature` of `TriMesh` would NOT return a triangle enclosing the point. This can happen when two triangles share an edge and that edge is closest to the query point.

This can happen because `project_local_point_and_get_feature` does not treat `TriMesh` as solid. However, in 2D TriMeshes can be treated as solid.

This change introduced by this PR is a subtly breaking change, because earlier versions would project to one of the edges.
I think breaking this API "contract" is reasonable, because the earlier behavior was misleading, with the returned `PointProjection` telling that the point was not inside and the feature id being incorrect, and always being set to a Face variant anyway.

### DIFF
--- a/src/query/point/point_composite_shape.rs
+++ b/src/query/point/point_composite_shape.rs
@@ -61,10 +61,7 @@ impl PointQuery for TriMesh {
             return (proj, feature_id);
         }
 
-        #[cfg(feature = "dim3")]
-        let solid = false;
-        #[cfg(feature = "dim2")]
-        let solid = true;
+        let solid = cfg!(feature = "dim2");
 
         let mut visitor =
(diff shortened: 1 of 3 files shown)


src/query/point/point_composite_shape.rs
```diff
@@ -61,8 +61,13 @@ impl PointQuery for TriMesh {
             return (proj, feature_id);
         }
 
+        #[cfg(feature = "dim3")]
+        let solid = false;
+        #[cfg(feature = "dim2")]
+        let solid = true;
```
Ralith, line 67: nit:
```suggestion
        let solid = cfg!(feature = "dim2");
```

PR #206, opened 2024-06-16
Respect stop_on_penetration flag in ball-ball shape casting

alexpardes:
This PR fixes a bug where `cast_shapes_ball_ball` ignored the `stop_on_penetration` flag and always returned a hit for separating intersecting balls.


sebcrozet:
Thanks!

PR #208, opened 2024-06-17
deps: Update to bitflags 2.x

waywardmonkeys:
This also removes an old style macro_use / extern crate for bitflags and makes all uses use the same syntax.


waywardmonkeys:
Because of `rkyv`, a different approach must be taken: https://github.com/bitflags/bitflags/issues/348

sebcrozet:
Thanks!

PR #216, opened 2024-07-02
Custom TypedShape to reference dyn Shape instead

Neo-Zhixing:
The existing `TypedShape::Custom` contains a user-defined shape with a type identified by a number.

This number has questionable utility. Let's make it so that `TypedShape` references the custom shape with a `&dyn Shape` instead. The user may check for the actual custom shape type using downcast.

Please let me know if there are any reasons to keep this as a `u32` I might not be aware of.


src/shape/shape.rs
```diff
@@ -150,8 +150,55 @@ pub enum TypedShape<'a> {
     #[cfg(feature = "dim2")]
     #[cfg(feature = "std")]
     RoundConvexPolygon(&'a RoundConvexPolygon),
-    /// A custom user-defined shape with a type identified by a number.
-    Custom(u32),
+    /// A custom user-defined shape.
+    Custom(&'a dyn Shape),
```
Vrixyz, line 154: Can a `dyn Shape` serialize/deserialize  ?

src/query/contact_manifolds/contact_manifolds_workspace.rs
```diff
@@ -65,7 +65,7 @@ impl DeserializableWorkspaceData {
             DeserializableWorkspaceData::CompositeShapeShapeContactManifoldsWorkspace(w) => {
                 Some(ContactManifoldsWorkspace(Box::new(w)))
             }
-            DeserializableWorkspaceData::Custom(_) => None,
+            DeserializableWorkspaceData::Custom => None,
```
Vrixyz, line 68: Do I understand that custom shapes are (currently) not supported by serialization/deserialization ? (at least here specifically)

Neo-Zhixing: That is correct. And this is the same behavior as before.

sebcrozet (approved):
LGTM, thank you! The `u32` was supposed to be a unique id identifying the custom shape type so that some custom serialization/deserialization code could handle it. Though that custom ser/deser isn’t implemented yet and will probably not be implemented for some time.
So I think it’s is fine to remove the `u32` for now. Eventually, it will probably be re-added but as a `Uuid` instead.

src/shape/shape.rs
sebcrozet: @Vrixyz This is difficult due to the `Deserialize` trait bound having a lifetime. This is generally worked around using `erased_serde` but some additional steps are needed to make deserialization actually work. This sounds out of the scope of this PR.

PR #217, opened 2024-07-02
Add `clone_scaled` to Shape traits

Neo-Zhixing:
We currently have some generic logic for scaling `Shape` in `bevy_rapier`. It seems reasonable to put this in the `Shape` trait so that the same logic can be reused elsewhere.

### DIFF
--- a/src/shape/shape.rs
+++ b/src/shape/shape.rs
@@ -344,7 +344,7 @@ pub trait Shape: RayCast + PointQuery + DowncastSync {
     fn clone_dyn(&self) -> Box<dyn Shape>;
 
     /// Scales this shape by `scale` into a boxed trait-object.
-    /// 
+    ///
     /// In some cases, the resulting shape doesn’t have the same type as Self. For example,
     /// if a non-uniform scale is provided and Self as a [`Ball`], then the result will be discretized
     /// (based on the `num_subdivisions` parameter) as a [`ConvexPolyhedron`] (in 3D) or [`ConvexPolygon`] (in 2D).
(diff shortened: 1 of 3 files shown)


src/shape/shape.rs
```diff
@@ -282,6 +282,9 @@ pub trait Shape: RayCast + PointQuery + DowncastSync {
     #[cfg(feature = "std")]
     fn clone_box(&self) -> Box<dyn Shape>;
 
+    #[cfg(feature = "std")]
+    fn clone_scaled(&self, scale: &Vector<Real>, num_subdivisions: u32) -> Option<Box<dyn Shape>>;
```
sebcrozet, line 286: This is lacking some documentation. The documentation should mention the fact that, in some cases, the resulting shape doesn’t have the same type as `Self`. For example, if a non-uniform scale is provided and `Self` as a `Ball`, then the result will be discretized (based on the `num_subdivisions` parameter) as a `ConvexPolyhedron` (in 3D) or `ConvexPolygon` (in 2D).

Because of that potential type change, I’m not sure that `clone_scaled` is the best name here (since `Clone` typically return the same type). Does calling it just `scaled` clash with the inherent functions `scaled` in any way when the users calls it?

sebcrozet (changes requested):
Thank you for this PR! Having this in parry directly sounds convenient.

src/shape/shape.rs
Neo-Zhixing: I've updated the docs.

I really don't want to create a naming ambiguity like that. Let's name it `dyn_scaled` instead then.

sebcrozet (approved):

src/shape/shape.rs
sebcrozet: Thank you. I did one last bit of renaming: `clone_box` is now `clone_dyn` and `dyn_scaled` is `scale_dyn`. So that the prefix/suffix are consistent.

sebcrozet:
Thanks!

PR #218, opened 2024-07-02
Composable QueryDispatcher

### LINKED ISSUE #8: Dispatch traits aren't composable
When processing queries on a compound shape, `QueryDispatcher` methods must be invoked recursively. This makes it infeasible for custom shapes that occur inside compound shapes to be handled. Ideally downstream code could provide `chain`able `QueryDispatcher` impls that only handle the shapes they introduce and otherwise return `Unsupported`, but in that case a compound shape would fall through to the default dispatcher, which would recurse internally and fail to handle custom shapes it encounters. This could be fixed by adding a `root_dispatcher: &dyn QueryDispatcher` argument to every trait method. The same issue also affects `PersistentQueryDispatcher`.

I can bang this out, but it's a bunch of boilerplate updates and given that this logic was not preserved from ncollide I want to be sure it's welcome before proceeding.

PixelDust22:
Resolves #8 

Implements @Ralith 's suggestion regarding composable `QueryDispatcher`.

> End-user ergonomics could be preserved by relegating the root_dispatcher argument to a trait for use only by shape implementers, and turning `QueryDispatcher` into a blanket-implemented trait that encapsulates the self.query(self, ...) dance, and provides a chain helper for composition.

I prefer this solution over the one suggested by @sebcrozet because @Ralith 's solution allows external crates to composably define shapes that need to dispatch recursively.

### DIFF
--- a/src/query/default_query_dispatcher.rs
+++ b/src/query/default_query_dispatcher.rs
@@ -1,15 +1,13 @@
 use crate::math::{Isometry, Point, Real, Vector};
 use crate::query::details::ShapeCastOptions;
 use crate::query::{
-    self, details::NonlinearShapeCastMode, ClosestPoints, Contact, NonlinearRigidMotion,
-    QueryDispatcher, ShapeCastHit, Unsupported,
+    self, details::NonlinearShapeCastMode, query_dispatcher::QueryDispatcherComposite,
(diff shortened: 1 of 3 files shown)


Ralith:
Thanks for building this out! This does look like what I had in mind. I didn't pursue it because I didn't have a pressing need, but I do think something like this is the right long term solution.

Neo-Zhixing:
@sebcrozet Can you give this a review?

PR #219, opened 2024-07-08
explore examples using bevy

Vrixyz:
An experiment to see how bevy could be used as a rendering backend for examples.

The code can be quite similar to macroquad's when using gizmos.

The big downside of using bevy is that compilation time is 2.5x slower (your mileage may vary), see attached timings (~40 seconds with macroquad, ~110 seconds with bevy) :

[timings.zip](https://github.com/user-attachments/files/16131361/timings.zip)

I think it makes sense to keep macroquad while we don't need more advanced features.

If/when we need more, it's encouraging that we can migrate to bevy rather easily.


Vrixyz:
We'll keep using macroquad for now, as using bevy both involves additional effort for readers (having a sense of ECS paradigm + bevy) and a cost of compilation speed... for not much benefit.

We'll reconsider that if a real need arises. To be noted macroquad is compatible with egui, if we ever need some accessible UI.

PR #221, opened 2024-07-08
Simplified mesh intersection logic

Makogan:
Change the mesh intersection code to rely on epsilon checks and point insertion to construct the final intersected mesh.


Makogan:
I know there are merge conflicts but I don;t know if we want to merge to master or to a different branch. So I await further instructions.

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -256,380 +355,358 @@ fn extract_connected_components(
...
+
+        if cos < worst_cos {
+            worst_cos = cos.abs();
         }
-        FeatureId::Vertex(vid) => FeatureId::Vertex(mesh.indices()[tri as usize][vid as usize]),
-        FeatureId::Face(_) => FeatureId::Face(tri),
-        FeatureId::Unknown => FeatureId::Unknown,
     }
+
+    worst_cos.acos() * 180. / PI
```
sebcrozet, line 577: This should return radians, not degrees.

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -256,380 +355,358 @@ fn extract_connected_components(
...
+
+    let mut worst_cos = 2.0;
+    for i in 0..points.len() {
+        let d1 = (points[i] - points[(i + 1) % n]).normalize();
+        let d2 = (points[(i + 2) % n] - points[(i + 1) % n]).normalize();
+
+        let cos = d1.dot(&d2);
+
+        if cos < worst_cos {
+            worst_cos = cos.abs();
```
sebcrozet, line 573: Is it normal that `cos < worst_cos` doesn’t take the absolute value of `cos`?

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -256,380 +355,358 @@ fn extract_connected_components(
...
+
+    (e1, e2)
+}
+
+fn project_point_to_segment(point: &Vector3<f64>, segment: &[Vector3<f64>; 2]) -> Vector3<f64> {
+    let dir = segment[1] - segment[0];
+    let local = point - segment[0];
+
+    let norm = dir.norm();
+    // restrict the result to the segment portion of the line.
```
sebcrozet, line 595:
```suggestion
    // Restrict the result to the segment portion of the line.
```

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -256,380 +355,358 @@ fn extract_connected_components(
...
+
+    segment[0] + coeff * dir.normalize()
+}
+
+/// No matter how smart we are about computing intersections. It is always possible
+/// to create ultra thin triangles when a point lies on an edge of a tirangle. These
+/// are degenerate and need to be terminated with extreme prejudice.
+fn is_triangle_degenerate(
+    triangle: &[Vector3<f64>; 3],
+    epsilon_degrees: f64,
```
sebcrozet, line 606: Should use radians.

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -256,380 +355,358 @@ fn extract_connected_components(
...
+    let norm = dir.norm();
+    // restrict the result to the segment portion of the line.
+    let coeff = (dir.dot(&local) / norm).clamp(0., norm);
+
+    segment[0] + coeff * dir.normalize()
+}
+
+/// No matter how smart we are about computing intersections. It is always possible
+/// to create ultra thin triangles when a point lies on an edge of a tirangle. These
+/// are degenerate and need to be terminated with extreme prejudice.
```
sebcrozet, line 603:
```suggestion
/// are degenerate and need to be removed.
```

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -256,380 +355,358 @@ fn extract_connected_components(
...
+        let proj = (triangle[i] - triangle[(i + 2) % 3]).dot(&dir) * dir + triangle[(i + 2) % 3];
+
+        worse_projection_distance = worse_projection_distance.min((proj - triangle[i]).norm());
+    }
+
+    if worse_projection_distance < epsilon_distance {
+        return true;
+    }
+
+    false
```
sebcrozet, line 638:
```suggestion
    worse_projection_distance < epsilon_distance
```

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -256,380 +355,358 @@ fn extract_connected_components(
...
-    mesh2: &TriMesh,
     flip2: bool,
-    spade_handle_to_intersection: &[HashMap<(u32, FixedVertexHandle), (FeatureId, FeatureId)>; 2],
-    intersection_points: &HashMap<(FeatureId, FeatureId), [Point<Real>; 2]>,
-    triangulations1: &HashMap<u32, Triangulation>,
-    triangulations2: &HashMap<u32, Triangulation>,
-    new_vertices12: &mut Vec<Point<Real>>,
-    new_indices12: &mut Vec<[u32; 3]>,
+    metadata: &MeshIntersectionMetadata,
+    mut point_set: &mut RTree<TreePoint>,
```
sebcrozet, line 648: Is the `mut` necessary?

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -256,380 +355,358 @@ fn extract_connected_components(
...
+
+        for face in delaunay.inner_faces() {
+            let verts = face.vertices();
+            let p1 = points[verts[0].index()];
+            let p2 = points[verts[1].index()];
+            let p3 = points[verts[2].index()];
+
+            // Sometimes the triangulation is messed up due to numerical errors. If
+            // a triangle does not survive this test. You can bet it should be put out
+            // of its misery.
```
sebcrozet, line 673:
```suggestion
            // a triangle does not survive this test, it should be ignored.
```

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -256,380 +355,358 @@ fn extract_connected_components(
...
+    let coeff = (dir.dot(&local) / norm).clamp(0., norm);
+
+    segment[0] + coeff * dir.normalize()
+}
+
+/// No matter how smart we are about computing intersections. It is always possible
+/// to create ultra thin triangles when a point lies on an edge of a tirangle. These
+/// are degenerate and need to be terminated with extreme prejudice.
+fn is_triangle_degenerate(
+    triangle: &[Vector3<f64>; 3],
```
sebcrozet, line 605: This should take points since they are the triangle’s vertices.
```suggestion
    triangle: &[Point3<f64>; 3],
```

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -256,380 +355,358 @@ fn extract_connected_components(
...
-                        let ins = *intersection_points
-                            .entry(key)
-                            .or_insert([inter.p1, inter.p2]);
-
-                        let _ = insert_point(ins, key, orig_fid, 0);
-                        let _ = insert_point(ins, key, orig_fid, 1);
-                    }
-                }
+fn insert_into_set(
+    position: Vector3<f64>,
```
sebcrozet, line 534:
```suggestion
    position: Point3<f64>,
```

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -256,380 +355,358 @@ fn extract_connected_components(
...
+    }
+
+    for e in edges {
+        writeln!(file, "l {} {}", e[0] + 1, e[1] + 1).unwrap();
+    }
+}
+
+#[derive(Copy, Clone, PartialEq, Debug, Default)]
+struct TreePoint {
+    point: Vector3<f64>,
```
sebcrozet, line 499: A field named `point` surely has to be a `Point3`, not a vector.

sebcrozet:
> I know there are merge conflicts but I don;t know if we want to merge to master or to a different branch. So I await further instructions.

@Makogan Yes, we want to merge to `master`.

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -256,380 +355,358 @@ fn extract_connected_components(
...
+                name: "".to_string(),
+                index: 0,
+                material: None,
+            }],
+            name: "".to_string(),
+        }],
+        ..Default::default()
+    }
+    .write_to_buf(&mut file)
+    .unwrap();
```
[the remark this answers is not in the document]
Makogan: I am not fond of returning the ObjData, the way the obj crate is structured has many flaws. I used it because it was the best I could find but in a perfect world I would have preferred something that separates the geometry data and meta-data information instead of nesting structures into structures. 

My suggestion is to just return an array of positions and indices for now.

src/transformation/mesh_intersection/mesh_intersection.rs
Makogan: It is, otherwise the code does not compile on my local.

src/transformation/mesh_intersection/mesh_intersection.rs
Makogan: A comment on the three points related to this discussion:

Mathematically, a point is a vector and a vector is a point. Having a distinction between the two increases the total code surface and amount of conversions that maintainers and users have to do. 

For example
```rs
fn smallest_angle(points: &[Point3<f64>]) -> f64 {
    let n = points.len();

    let mut worst_cos = 2.0;
    for i in 0..points.len() {
        let d1 = (points[i].coords - points[(i + 1) % n].coords).normalize();
        let d2 = (points[(i + 2) % n].coords - points[(i + 1) % n].coords).normalize();

        let cos = d1.dot(&d2);

        if cos < worst_cos {
            worst_cos = cos.abs();
        }
    }

    worst_cos.acos() * 180. / PI
}
```

Is adding `.coords` here helping make the code more explicit, or is it adding more effort for the programmer and more complexity for the reader?

Given how common a person needs to compute different kinds of inner products, projections, translations, orthogonalizations... It feels like the `Point` interface and its traits associations produce more difficulties than benefits.

I will change things to comply with the current standard. But I leave these comments to express my humble personal opinion of something that I believe might cause more harm than good.

src/transformation/mesh_intersection/mesh_intersection.rs
Makogan: A point is a vector! But I will change it T.T

Makogan:
Merged to master and made sure the current code passes the CI checks

src/transformation/mesh_intersection/mesh_intersection.rs
sebcrozet: The distinction between points and vectors is mainly to avoid mistakes, like if you wrote `point[i].dot(&point[i + 1])` which would obviously not make sense because that’s taking the dot product of two location in space.

I recommend reading some resources about that distinction. Searching for the distinction between points and vectors yields many results. To site one, from [the university of california](https://ed.iitm.ac.in/~raman/Points-and-Vectors.pdf):
> These objects are often confused when studying the field and frequently students are taught that points and vectors are “essentially” the same. However, they are very different. Namely,
> • A Point has position in space. The only characteristic that distinguishes one point from another is its
position.
> • A Vector has both magnitude and direction, but no fixed position in space.

Regarding your example:


> Is adding .coords here helping make the code more explicit, or is it adding more effort for the programmer and more complexity for the reader?

Adding `.coords` here only illustrates a misunderstanding of the concept of points. Since points are location in spaces, and a vector the displacement between two locations, then the subtraction of two points is a vector. So you can just write `(points[i]- points[(i + 1) % n]).normalize()` because `points[i]- points[(i + 1) % n]` is already a vector in the first place.

Having a distinct type between points and vectors has some additional perks when combined with transformations. For example if `iso` is an isometry, `iso * vector` will only rotate the vector whereas `iso * point` will both rotate and translate it. This prevents a wide range of errors where you would forget to apply the translational part to a vector.

src/transformation/mesh_intersection/mesh_intersection.rs
sebcrozet: Sorry, but a point is just not the same as a vector. The fact that you called the variable `point` instead of `vector` shows that, through the name of the argument, you wanted to give semantic to that vector. A `Point3` gives you that semantic at the type level.

src/transformation/mesh_intersection/mesh_intersection.rs
Makogan: The resources that talk about that distinction usually come from the more physics side of mathematics. Many resources will define points and vectors to be the same, for example in wolfram:

> In dimensions greater than or equal to two, points are sometimes considered synonymous with [vectors](https://mathworld.wolfram.com/Vector.html) and so points in [n-dimensional space](https://mathworld.wolfram.com/n-Space.html) are sometimes called [n-vectors](https://mathworld.wolfram.com/n-Vector.html).
https://mathworld.wolfram.com/Point.html

> point[i].dot(&point[i + 1]) which would obviously not make sense

I mean, a contrived counter example is if the points are on a circle and we are looking for the angle between them or if we want to know the winding number of the origin relative to a poly line. 

The only reason why we do things like  `points[i]- points[(i + 1) % n]` is to move objects to the origin to then apply the dot product. So technically speaking `point[i].dot(&point[i + 1])` is the only actually valid operation when taking the dot product, everything else we do is just trying to align our space with the correct origin so that our calculations make sense.

> Since points are location in spaces, and a vector the displacement between two locations,

A vector is not a displacement, a vector is the linear combination of a basis. And since points are represented as coordinates, they are, implicitly, a linear combination of an orthonormal basis.

The idea that vectors are displacements is a specific interpretation in the context of physics, but for example, polynomials and matrices are vector spaces and neither represents a displacement.

This is coming more to a matter of opinion and rather pedantic, as I stated at the beginning. I wish not to distract you from more important subjects. I just wanted to give you my two cents, please disregard this discussion.

src/transformation/mesh_intersection/mesh_intersection.rs
Makogan: See my comment above. I called it a point because the examples on the library I used for the rtree named it point and I just copy pasted that example and edited accordingly.

On the latest branch, all variables have been renamed according to the specification, i.e. they are all `Point3`.

crates/parry3d-f64/Cargo.toml
```diff
@@ -59,6 +60,11 @@ cust_core = { version = "0.1", optional = true }
 spade = { version = "2", optional = true } # Make this optional?
 rayon = { version = "1", optional = true }
 bytemuck = { version = "1", features = ["derive"], optional = true }
+rstar = "0.12.0"
+obj = { version = "0.10.2", optional = true }
+
+[target.'cfg(not(target_os = "cuda"))'.dependencies]
+cust = { version = "0.3", optional = true }
```
sebcrozet, line 67: This should be removed. This is probably an artifact introduced by the merge with `main`.
```suggestion
```

src/shape/trimesh.rs
```diff
@@ -172,18 +172,25 @@ pub struct TriMeshTopology {
...
 
-        result
+#[cfg(all(feature = "std", feature = "cuda"))]
+impl TriMeshTopology<DefaultStorage> {
+    fn to_cuda(&self) -> CudaResult<TriMeshTopology<CudaStorage>> {
+        Ok(TriMeshTopology {
+            vertices: CudaArray1::new(&self.vertices)?,
+            faces: CudaArray1::new(&self.faces)?,
+            half_edges: CudaArray1::new(&self.half_edges)?,
+        })
```
sebcrozet, line 193: These should be removed. Looks like another merge artifact (there is no cuda support anymore).

src/transformation/mesh_intersection/triangle_triangle_intersection.rs
```diff
@@ -250,6 +175,40 @@ pub fn triangle_triangle_intersection(
     }
 }
 
+/// Smarter, but more expensive, mechanism to find a triangle normal.
+fn robust_triangle_normal(tri: &Triangle) -> na::Vector3<Real> {
```
sebcrozet, line 179: This should be a method in the `Triangle` struct. I’ve made [a comment](https://github.com/dimforge/parry/pull/221#discussion_r1674206434) in the previous review but that likely got unnoticed.

src/transformation/wavefront.rs
```diff
@@ -0,0 +1,42 @@
+use crate::shape::TriMesh;
+
+#[cfg(feature = "wavefront")]
+use obj::*;
+#[cfg(feature = "wavefront")]
+use std::path::PathBuf;
+
+#[cfg(feature = "wavefront")]
```
sebcrozet, line 8: This are unnecessary because the `pub mod wavefront;` itself is already feature-gated.
```suggestion
use obj::*;
use std::path::PathBuf;
```

src/transformation/wavefront.rs
```diff
@@ -0,0 +1,42 @@
+use crate::shape::TriMesh;
+
+#[cfg(feature = "wavefront")]
+use obj::*;
```
sebcrozet, line 4: The glob-import should be replaced by importing each relevant member explicitly.

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -256,393 +348,551 @@ fn extract_connected_components(
...
+
+        if cos < worst_cos {
+            worst_cos = cos.abs();
         }
-        FeatureId::Vertex(vid) => FeatureId::Vertex(mesh.indices()[tri as usize][vid as usize]),
-        FeatureId::Face(_) => FeatureId::Face(tri),
-        FeatureId::Unknown => FeatureId::Unknown,
     }
+
+    worst_cos.acos() * 180. / PI as Real
```
sebcrozet, line 537: Looks like my previous review’s comment about sticking to radians went unnoticed.

sebcrozet (changes requested):
Thank you for this first round of changes.
I noticed that a few remarks went unnoticed. Don’t forget that github tends to collapse multiple conversations if there are many of them.

Two other remarks:
- The obj test files should be moved to a `assets/tests` directory at the root of the repository.
- Please add authoring & licence information as a comment in the `.obj` files.

src/shape/trimesh.rs
```diff
@@ -239,21 +239,6 @@ impl TriMeshTopology<CudaStorage> {
...
-        let mut result = [first_half_edge; 3];
-        for k in 1..3 {
-            let half_edge = self.half_edges[result[k - 1] as usize];
-            result[k] = half_edge.next;
-        }
-
-        result
-    }
-}
-
```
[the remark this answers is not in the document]
Makogan: The CI returns an error because clippy complaints that that code is not used anywhere. So i deleted it to pass the check.

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -1,11 +1,45 @@
...
+    /// a smaller angle is considered degenerate and will be deleted.
+    pub angle_epsilon: f64,
+    /// The maximum distance at which two points are considered to overlap in space
+    /// if `||p1 - p2|| < global_insertion_epsilon` then p1 and p2 are considered
+    /// to be the same point.
+    pub global_insertion_epsilon: f64,
+    /// A multiplier coefficient to scale `global_insertion_epsilon` when checking for
+    /// point duplicatin within a single triangle. Inside of an individual triangle
+    /// the distance at wich two points are considered to be the same is
+    /// `global_insertion_epsilon * local_insertion_epsilon_mod`.
```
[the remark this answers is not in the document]
Makogan: modifier, it's a multiplicative modifier. It was already quite long, which is why I shortened it.

Makogan:
Other than the conversations marked as unresolved, I think I addressed all comments.

src/transformation/mesh_intersection/mesh_intersection.rs
Makogan: You are right that there is something wonky here, but it is not the absolute value. We are looking for the smallest angle, so the largest cos. The actual logic should have been grabbing the maximum of all unsigned cos. Thank you for catching this, I was quite lucky the algorithm was working despite the bug.

```rs
fn smallest_angle(points: &[Point3<Real>]) -> Real {
    let n = points.len();

    let mut worst_cos = -2.0;
    for i in 0..points.len() {
        let d1 = (points[i].coords - points[(i + 1) % n].coords).normalize();
        let d2 = (points[(i + 2) % n].coords - points[(i + 1) % n].coords).normalize();

        let cos = d1.dot(&d2);

        if cos > worst_cos {
            worst_cos = cos;
        }
    }

    worst_cos.acos()
}
```

Makogan: The problem is actually that I should have been looking for the largest signed `cos`. Very lucky the code works despite this bug.

```rs
fn smallest_angle(points: &[Point3<Real>]) -> Real {
    let n = points.len();

    let mut worst_cos = -2.0;
    for i in 0..points.len() {
        let d1 = (points[i].coords - points[(i + 1) % n].coords).normalize();
        let d2 = (points[(i + 2) % n].coords - points[(i + 1) % n].coords).normalize();

        let cos = d1.dot(&d2);
        if cos > worst_cos {
            worst_cos = cos;
        }
    }

    worst_cos.acos()
}
```

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -256,393 +348,547 @@ fn extract_connected_components(
...
+    let norm = dir.norm();
+    // Restrict the result to the segment portion of the line.
+    let coeff = (dir.dot(&local) / norm).clamp(0., norm);
+
+    segment[0] + coeff * dir.normalize()
+}
+
+/// No matter how smart we are about computing intersections. It is always possible
+/// to create ultra thin triangles when a point lies on an edge of a tirangle. These
+/// are degenerate and need to be terminated with extreme prejudice.
```
sebcrozet, line 564: Looks like my previous comment marked as resolved without being applied?
```suggestion
/// are degenerate and need to be removed.
```

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -256,393 +348,547 @@ fn extract_connected_components(
...
+    let mut worse_projection_distance = Real::MAX;
+    for i in 0..3 {
+        let dir = triangle[(i + 1) % 3] - triangle[(i + 2) % 3];
+        if dir.norm() < epsilon_distance {
+            return true;
+        }
+
+        let dir = dir.normalize();
+        let proj =
+            (triangle[i] - triangle[(i + 2) % 3]).dot(&dir) * dir + triangle[(i + 2) % 3].coords;
```
sebcrozet, line 591: The `.coords` can be avoided by writing this as a `Point + Vector` operation.
```suggestion
        let proj =
            triangle[(i + 2) % 3] + (triangle[i] - triangle[(i + 2) % 3]).dot(&dir) * dir;
```

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -256,393 +348,547 @@ fn extract_connected_components(
...
+
+        worse_projection_distance =
+            worse_projection_distance.min((proj - triangle[i].coords).norm());
```
sebcrozet, line 594: If the suggestion above is applied, this `.coords` isn’t needed either.
```suggestion
            worse_projection_distance.min((proj - triangle[i]).norm());
```

sebcrozet (changes requested):
Thank you for this second round of changes.
There are still a few ones left and a couple I haven’t seen the first times.

Regarding the obj files: please move them to `assets/tests` instead of `test_data`.

src/shape/shape.rs
```diff
@@ -1400,3 +1402,28 @@ impl_shape_for_round_shape!(
...
+
+        let cos_abs = cos.abs();
+        if cos_abs < best_cos {
+            best_cos = cos_abs;
+            selected_i = i;
+        }
+    }
+
+    selected_i
+}
```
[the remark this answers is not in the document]
Makogan: It is also used in the mesh intersection code inside of `triangulate_constraints_and_merge_duplicates`, line 407. That's why i had to move it there. So that I had access to it.

sebcrozet (approved):

sebcrozet:
Thank you, I took care of the last few details.

PR #223, opened 2024-07-15
Replace type alias with a 'use'.

### LINKED ISSUE #222: Migrate to "use" instead of alias pattern for `Real`
As a new user opening rapier, it can be overwhelming to see `Real` used everywhere, rapier uses heavily generic code, so thriving to limit indirection can help.

```rs
// This can be gated behind a feature flag for f64 or f32.
use f64 as Real;

pub fn takes_a_real(param: Real) {}
```

use_pattern_doc

Idea credit to Alice from bevy ; and https://www.reddit.com/r/rust/comments/1e0trid/comment/lcpjwh4.

Vrixyz:
Fixes #222

image


src/lib.rs
```diff
@@ -77,11 +77,11 @@ pub mod utils;
 mod real {
     /// The scalar type used throughout this crate.
     #[cfg(feature = "f64")]
-    pub type Real = f64;
+    pub use f64 as Real;
```
Vrixyz, line 80: a changelog line would probably be welcome.

sebcrozet (approved):
Good find, thanks! 

We should consider doing that for the other aliases, e.g. `use na::Vector3 as Vector` instead of `type Vector<N> = Vector3<N>` (can be in a separate PR).

+1 for adding a changelog line.

Vrixyz:
> We should consider doing that for the other aliases, e.g. `use na::Vector3 as Vector` instead of `type Vector<N> = Vector3<N>` (can be in a separate PR).

- See https://github.com/dimforge/parry/pull/232 
  - the difference is more subtle and I'm not sure it's more consistent, it still redirects to an alias (which is alright, but has this big dislaimer **"Because this is an alias, not all its methods are listed here. See the Point type too."** ; but yes let's discuss it in the other PR.

Vrixyz:
This adds a warning when running `cargo +nightly clippy`:

```
warning: casting to the same type is unnecessary (`f32` -> `f32`)
    --> crates\parry3d\../../src\shape\shape.rs:1466:9
     |
1466 |         f32::MAX as Real
     |         ^^^^^^^^^^^^^^^^ help: try: `f32::MAX`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_cast
```

Interestingly, this first one seems correct ^ ; but the following is more difficult to fix:

```
warning: casting float literal to `f32` is unnecessary
  --> crates\parry3d\../../src\query\sat\sat_cuboid_cuboid.rs:14:18
   |
14 |     let signum = (1.0 as Real).copysign(pos12.translation.vector.dot(axi... 
   |                  ^^^^^^^^^^^^^ help: try: `1.0_f32`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_cast
```

waywardmonkeys:
`Real::MAX` was what I intended to do when I got to these.

For the other, you could try `Real::from(1.0)` and see if that warns ... or a quick trait like the `One` trait from `num`.

PR #224, opened 2024-07-15
cargo doc ci

Vrixyz:


waywardmonkeys:
I hadn't seen this when I pushed my #226 and #227 ... my approach in #227 is slightly different, so it is worth comparing them.

.github/workflows/parry-ci-build.yml
```diff
@@ -22,6 +22,18 @@ jobs:
...
+  doc:
+    runs-on: ubuntu-latest
+    env:
+      RUSTDOCFLAGS: -D warnings
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+        with:
+          prefix-key: ${{ env.RUST_CACHE_KEY }}
```
Vrixyz, line 34: I added that for consistency with rapier, but it might not be necessary 🤔 , I like the more straightforward approach of https://github.com/dimforge/parry/pull/227

sebcrozet (approved):
Thanks! I approved it, but feel free to merge #227 (or a combination of the two) if that sounds better.
I agree that compiling with some features enabled is more robust.

sebcrozet:
#224 got merged so I believe we can close this one. Feel free to reopen otherwise.

PR #227, opened 2024-07-17
ci: Add `build-doc` job to test building docs.

waywardmonkeys:
This doesn't use nightly (like docs.rs does) because there are other warnings on nightly builds that would break this.

Also, I used `parallel` here but that isn't tested yet by other build configs.

### DIFF
--- a/.github/workflows/parry-ci-build.yml
+++ b/.github/workflows/parry-ci-build.yml
@@ -67,3 +67,15 @@ jobs:
         run: xargo build --verbose --no-default-features --features required-features --target=x86_64-unknown-linux-gnu;
       - name: build thumbv7em-none-eabihf
         run: xargo build --verbose --no-default-features --features required-features --target=thumbv7em-none-eabihf;
+  build-doc:
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
This will fail until #226 is landed.

.github/workflows/parry-ci-build.yml
```diff
@@ -67,3 +67,15 @@ jobs:
...
+    env:
+      RUSTDOCFLAGS: -D warnings
+    steps:
+      - uses: actions/checkout@v4
+      - name: install stable Rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+            toolchain: stable
+      - name: cargo doc
+        run: cargo doc --workspace --features bytemuck-serialize,serde-serialize,rkyv-serialize,parallel --no-deps --document-private-items
```
Vrixyz, line 81: features addition is a good addition compared to https://github.com/dimforge/parry/pull/226

Vrixyz:
> This will fail until #226 is landed.

and https://github.com/dimforge/parry/pull/235 too.

sebcrozet:
Thank you! I rebased on top of `master` after merging your other fixes. All green now ✅

PR #231, opened 2024-07-17
ci: Use a regular `cargo` to build no-std targets.

waywardmonkeys:
`xargo` is no longer needed and has been in maintenance mode for over 7 years.

This can run on `stable` rather than requiring `nightly`

The build doesn't need `rustfmt`, so don't require that it be installed.


waywardmonkeys:
This will conflict with #227, but I'm happy to fix up any merge conflicts after whichever lands first.

Vrixyz (approved):
I'm approving, but I have a minor concern over dropping nightly compilation: that means we don't have any ci testing nightly.

It's most definitely correct for this PR, but should we add a periodic CI step on master to test against nightly ? @sebcrozet

sebcrozet:
Thank you for this PR!

Back in the days, `xargo` was the only way to ensure that compilation actually fails if any crate (including indirect dependencies) attempts to link against the stdlib. Is that what we gain from `cargo` now as well? i.e. if we add some crate that needs the stdlib to parry’s `Cargo.toml`, will it fail to compile when targetting `thumbv7em-none-eabihf`?

> It's most definitely correct for this PR, but should we add a periodic CI step on master to test against nightly ?

@Vrixyz We should keep a nightly test (on PRs) that checkes the simd-nightly feature. (But, yes, this should be in a separate PR.)

waywardmonkeys:
> Back in the days, `xargo` was the only way to ensure that compilation actually fails if any crate (including indirect dependencies) attempts to link against the stdlib. Is that what we gain from `cargo` now as well? i.e. if we add some crate that needs the stdlib to parry’s `Cargo.toml`, will it fail to compile when targetting `thumbv7em-none-eabihf`?

Yes.

Using `cargo` with a target that has no stdlib is (now) the typical way to verify that `no_std` is indeed working correctly.

This is what we do in `kurbo` for example as well:

https://github.com/linebender/kurbo/blob/16cb9af39eda5010ec30eb086822f741edfdf9f5/.github/workflows/ci.yml#L50

waywardmonkeys:
As an aside, this would be a great time to bring up the work started in PR #170 and see about finding a victim to do more of that. :)

Vrixyz:
Is this related to the failure I'm seeing in https://github.com/dimforge/parry/pull/249 ? :eyes:

sebcrozet (approved):
Thank you for checking @waywardmonkeys! Looks good.

PR #232, opened 2024-07-17
Explore more controversial "use"s for types

Vrixyz:
<details><summary>Details</summary>
<p>

This pull request:

image

Existing (https://docs.rs/parry3d/0.16.1/parry3d/math/type.Point.html):

(note there is no scroll there, there is way less information.)

image

I'm not a huge fan of the big disclaimer **"Because this is an alias, not all its methods are listed here. See the Point type too."** 💭 

</p>
</details> 

❓ I'm surprised `Vector` shows as `Vector` rather than `Vector2` (or `Vector3`).

- and that the alias points to `na::Matrix` rather than `na::Vector`.

- given this :
```rs
pub use Vector2 as Vector;
```
image
- I'm surprised that it doesn't render as `Vector2` ; where `Real` correctly render as `f32`
image
- and directs to
- image

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,11 +77,11 @@ pub mod utils;
 mod real {
     /// The scalar type used throughout this crate.
     #[cfg(feature = "f64")]
-    pub type Real = f64;
+    pub use f64 as Real;
 
     /// The scalar type used throughout this crate.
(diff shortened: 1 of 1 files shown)


Vrixyz:
Closing this as I didn´t succeed in convincing results.

PR #235, opened 2024-07-18
Fix rustdoc error in `Triangle`.

waywardmonkeys:
Commit 4ffa010f99a38ebd6d7ac763355403e98454fc09 introduced a method that is only available with feature = "dim3", but a method that is present when feature = "dim2" was linking to it. The doc commnets referencing it must be protected by `cfg_attr` checks.

### DIFF
--- a/src/shape/shape.rs
+++ b/src/shape/shape.rs
@@ -893,10 +893,13 @@ impl Shape for Triangle {
 
     fn feature_normal_at_point(
         &self,
-        feature: FeatureId,
+        _feature: FeatureId,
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
@Vrixyz This is needed so that one of #224 or #227 can land.

sebcrozet:
Thank you for this PR! The issue here is that the `Triangle::normal` and `::scaled_normal` method should only be available in 3D too.

waywardmonkeys:
@sebcrozet Well, that's a ton easier than what I did. :)

PR #237, opened 2024-07-21
Fix `unexpected_cfgs` warnings

waywardmonkeys:
This will add a one line warning from `cargo` until 1.80 (which is due for release soon).


waywardmonkeys:
This adds something to prevent `improved_fixed_point_support` from causing issues, but that code should be revisited and either a feature added for it or the code removed perhaps.

Vrixyz:
we'll probably want to merge that in to fix the rest of the warnings: https://github.com/waywardmonkeys/parry/pull/1

PR #238, opened 2024-07-21
Fix `clippy::doc_lazy_continuation` lints

waywardmonkeys:
These are new in Rust 1.80


src/shape/polyline.rs
```diff
@@ -218,6 +218,7 @@ impl Polyline {
     ///   `num_indices == self.indices.len()`.
     /// - This polyline is oriented counter-clockwise.
     /// - In 3D, the polyline is assumed to be fully coplanar, on a plane with normal given by
```
waywardmonkeys, line 220: In another commit, someone should fix this comment to complete the sentence to say what that normal is given by.

Vrixyz: https://github.com/dimforge/parry/pull/247

PR #243, opened 2024-07-23
Fix unlisted features

Vrixyz:
detected thanks to https://github.com/dimforge/parry/pull/237


crates/parry3d/Cargo.toml
```diff
@@ -30,6 +30,9 @@ simd-stable = ["simba/wide", "simd-is-enabled"]
 simd-nightly = ["simba/portable_simd", "simd-is-enabled"]
 enhanced-determinism = ["simba/libm_force", "indexmap"]
 parallel = ["rayon"]
+wavefront = ["obj"]
```
waywardmonkeys, line 33: This should be `“dep:obj”` …

waywardmonkeys: (But a lot of things like that need fixing … so a separate PR for those will be forthcoming tomorrow.)

Issue #205, opened 2024-04-28
EPA convergence failure on capsules


sebcrozet:
See the following PR for a code to reproduce the bug: https://github.com/dimforge/rapier/pull/531

jjanella:
Through the godot-rapier plugin this bug is causing frame drops in a game I am developing

Ughuuu:
The gjk returns intersection, and then epa2 starts.
Adding some logs, it seems it never converges:
```
Iteration 1: curr_dist = 7.9495697, candidate_max_dist = 11.553711 max_dist = 11.553711, support_point_id = 3
Iteration 2: curr_dist = 8.539658, candidate_max_dist = 9.600317 max_dist = 9.600317, support_point_id = 4
Iteration 3: curr_dist = 9.436261, candidate_max_dist = 9.436432 max_dist = 9.436432, support_point_id = 5
Iteration 4: curr_dist = 9.436262, candidate_max_dist = 9.436306 max_dist = 9.436306, support_point_id = 6
Iteration 5: curr_dist = 9.436289, candidate_max_dist = 9.43631 max_dist = 9.436306, support_point_id = 7
Iteration 6: curr_dist = 9.436289, candidate_max_dist = 9.43631 max_dist = 9.436306, support_point_id = 8
Iteration 7: curr_dist = 9.436289, candidate_max_dist = 9.43631 max_dist = 9.436306, support_point_id = 9
Iteration 8: curr_dist = 9.436289, candidate_max_dist = 9.43631 max_dist = 9.436306, support_point_id = 10
Iteration 9: curr_dist = 9.436289, candidate_max_dist = 9.43631 max_dist = 9.436306, support_point_id = 11
Iteration 10: curr_dist = 9.436289, candidate_max_dist = 9.43631 max_dist = 9.436306, support_point_id = 12
Iteration 11: curr_dist = 9.436289, candidate_max_dist = 9.43631 max_dist = 9.436306, support_point_id = 13
Iteration 12: curr_dist = 9.436289, candidate_max_dist = 9.43631 max_dist = 9.436306, support_point_id = 14
Iteration 13: curr_dist = 9.436289, candidate_max_dist = 9.43631 max_dist = 9.436306, support_point_id = 15
Iteration 14: curr_dist = 9.436289, candidate_max_dist = 9.43631 max_dist = 9.436306, support_point_id = 16
[...]

PR #245, opened 2024-07-27
Improve convergence of epa algorithm in case it's stuck

Ughuuu:
Improves converges by checking if it's stuck. In such a case return a result, as a better solution will not be found, and most likely they are intersecting but not within the EPS.

Also decrease max iterations from 10000 to 100. It usually converges in 4-5 iterations, no need to ever do as many. Also we never really want a precision that big, we are ok with a lower precision most times.

- Fixes https://github.com/dimforge/parry/issues/205

Possibly fixes:
- https://github.com/dimforge/parry/issues/157

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## Unreleased
+
+### Modified
+
+- Improve convergence of epa algorithm in degenerate configurations.
+
 ## v0.17.0
 
 ### Added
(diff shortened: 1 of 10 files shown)


Ughuuu:
With this fix it actually returns a contact (didn't know how to draw the intersection better):

image

Also it does so in 5 iterations, instead of doing 10000 iterations and returning None.

Vrixyz:
Thanks! I'd like to add a test for this scenario, would a setup like https://github.com/dimforge/rapier/pull/531 be the correct candidate ?

Ughuuu:
Sure, thats what I used locally.

Vrixyz:
:tada: yay! I'd like to add a changelog line before merging though

PR #248, opened 2024-07-29
Fix infinite loop in `intersection_with_local_plane`

### LINKED ISSUE #215: function hang infinitely
https://github.com/dimforge/parry/blob/b72ccd4f95ab54d3b71a743958ba91dc25727764/src/query/split/split_trimesh.rs#L584-L598

when none of condition it will hang infinitely

ThierryBerger:
- Attempt to fix https://github.com/dimforge/parry/issues/215

Only failing test for now

the loop occurs within the `'traversal` labelled loop, the 2 `if` conditions are never true in our case.

image

## Wild guess

It seems weird that we have twice the same index in the index_adjacencies list ? But I'm not sure there:

image

### DIFF
--- a/crates/parry3d/tests/geometry/trimesh_intersection.rs
+++ b/crates/parry3d/tests/geometry/trimesh_intersection.rs
@@ -121,3 +121,27 @@ fn trimesh_plane_below() {
 
     assert!(matches!(result, IntersectResult::Negative));
 }
+
+#[test]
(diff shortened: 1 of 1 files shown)


sebcrozet:
Thank you. Sorry for getting to this late. This seems to already be fixed on master now as I cannot reproduce it. Please reopen if that’s not the case.

PR #250, opened 2024-07-31
Add an example for projection

Vrixyz:
This Pr also adds some common code for macroquad examples

[Screencast from 09-03-2024 11:56:59 AM.webm](https://github.com/user-attachments/assets/8170b454-99ad-4696-9093-4f93fd6a5cb0)


Vrixyz:
the vertices seem incorrect:

See (left = current indices ; right = indices order after transforming into convex polyheron, which works fine):

image

src/transformation/to_trimesh/cuboid_to_trimesh.rs
```diff
@@ -28,36 +37,54 @@ impl Cuboid {
  *
  * The cuboid is centered at the origin, and has its half extents set to 0.5.
  */
-fn unit_cuboid() -> (Vec<Point3<Real>>, Vec<[u32; 3]>) {
-    let mut coords = Vec::with_capacity(8);
-    let mut faces = Vec::with_capacity(12);
+fn unit_cuboid() -> (Vec<PointDim<Real>>, Vec<[u32; 3]>) {
+    #[cfg(feature = "dim3")]
+    return {
```
sebcrozet, line 42: Not 100% sure, but, I don’t think you need the `return`? (Assuming the `;` after `}` is removed.)

src/transformation/to_trimesh/cuboid_to_trimesh.rs
```diff
@@ -2,11 +2,20 @@ use crate::bounding_volume::Aabb;
...
+#[cfg(feature = "dim3")]
+use na::Point3 as PointDim;
+#[cfg(feature = "dim3")]
 use na::{self, Point3};
 
+#[cfg(feature = "dim2")]
+use na::Point2 as PointDim;
+#[cfg(feature = "dim2")]
+use na::{self, Point2};
+
```
sebcrozet, line 15: You can remove these, and just use `crate::math::Point` instead of `PointDim/Point2/Point3`.

crates/parry3d/examples/project_point3d.rs
```diff
@@ -0,0 +1,91 @@
...
+
+        let elapsed_time = get_time() as f32;
+        let slow_elapsed_time = elapsed_time / 3.0;
+
+        let point_to_project = lissajous_3d(slow_elapsed_time);
+        let projected_point = trimesh.project_point(
+            &Isometry::identity(),
+            &na_from_mquad(point_to_project),
+            true,
+        );
```
sebcrozet, line 32: Since the transform is the identity, we can call `project_local_point` here:
```suggestion
        let projected_point = trimesh.project_local_point(
            &na_from_mquad(point_to_project),
            true,
        );
```

crates/parry3d/examples/project_point3d.rs
```diff
@@ -0,0 +1,91 @@
...
+
+        let slow_elapsed_time = slow_elapsed_time * 0.7;
+        // Going 3d!
```
sebcrozet, line 35:
```suggestion
        // Setup 3D camera.
```

sebcrozet (approved):
The point type system needs to be simplified in `cuboid_to_trimesh.rs`, but looking good otherwise. Thanks!

src/transformation/to_trimesh/cuboid_to_trimesh.rs
Vrixyz: yes that works :+1:

PR #255, opened 2024-09-02
Parry benchmarks compilation

Vrixyz:
Fixes compiling for parry benchmarks

There has been a few API changes since their creation, especially around `ConvexHull`, and in a lesser amount `Capsule`.

### DIFF
--- a/crates/parry3d/Cargo.toml
+++ b/crates/parry3d/Cargo.toml
@@ -89,10 +89,7 @@ oorandom = "11"
 ptree = "0.4.0"
 rand = { version = "0.8" }
 macroquad = "0.4.12"
-nalgebra = { version = "0.33", default-features = false, features = [
-    "libm",
-    "rand",
-] }
+nalgebra = { version = "0.33", default-features = false, features = ["rand"] }
 rand_isaac = "0.3"
 
 [package.metadata.docs.rs]
--- a/crates/parry3d/Cargo.toml
+++ b/crates/parry3d/Cargo.toml
@@ -89,6 +89,11 @@ oorandom = "11"
 ptree = "0.4.0"
 rand = { version = "0.8" }
 macroquad = "0.4.12"
+nalgebra = { version = "0.33", default-features = false, features = [
+    "libm",
+    "rand",
+] }
+rand_isaac = "0.3"
 
 [package.metadata.docs.rs]
 rustdoc-args = ["-Zunstable-options", "--generate-link-to-definition"]
--- a/crates/parry3d/benches/all.rs
+++ b/crates/parry3d/benches/all.rs
@@ -1,6 +1,8 @@
 #![feature(test)]
 #![allow(unused_macros)]
 
+use parry3d::math::{Point, Real};
+
 extern crate nalgebra as na;
 extern crate parry3d;
 extern crate rand;
@@ -10,3 +12,8 @@ mod bounding_volume;
 mod common;
 mod query;
 mod support_map;
+
+#[cfg(feature = "dim2")]
+type ConvexHull = Vec<Point<Real>>;
+#[cfg(feature = "dim3")]
+type ConvexHull = (Vec<Point<Real>>, Vec<[u32; 3]>);
--- a/crates/parry3d/benches/all.rs
+++ b/crates/parry3d/benches/all.rs
@@ -1,8 +1,6 @@
 #![feature(test)]
 #![allow(unused_macros)]
 
-use parry3d::math::{Point, Real};
-
 extern crate nalgebra as na;
 extern crate parry3d;
 extern crate rand;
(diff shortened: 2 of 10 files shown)


crates/parry3d/benches/bounding_volume/mod.rs
```diff
@@ -143,7 +142,7 @@ bench_method!(
     bounding_sphere: BoundingSphere,
     c: ConvexHull,
     m: Isometry3<f32>
-);
+);*/
```
Vrixyz, line 145: `ConvexHull` is no longer a type as iitself, but rather an alias to `Vec<Point<Real>>`, I think it would have value to be its own type, but slightly out of scope for this PR ; I'd appreciate instructions as how to proceed:

Should I straight up remove that code, or keep it in and create a tracking issue (+ comment) ?

Then, should I either:
- make it work with the alias ? 
  - implement the methods for `Vec<Point<Real>>`, which will probably be confusing
- make a `ConvexHull` type. An opportunity for better docs in my opinion, `ConvexHull` exists in the current documentation at a few places ([here](i.e: https://github.com/Vrixyz/parry/blob/7e76ebe0c7a155e44d631d4f25a227cf7daf598c/src/query/closest_points/closest_points_support_map_support_map.rs#L8) or [there](https://github.com/Vrixyz/parry/blob/7e76ebe0c7a155e44d631d4f25a227cf7daf598c/src/query/intersection_test/intersection_test_halfspace_support_map.rs#L16-L17)...)

crates/parry3d/benches/support_map/mod.rs
```diff
@@ -59,10 +59,12 @@ bench_method!(
...
 );
+/*
 bench_method!(
     bench_convex_support_map,
     support_point,
     c: ConvexHull,
     m: Isometry3<f32>,
     dir: Vector3<f32>
 );
+*/
```
Vrixyz, line 70: see other comment: https://github.com/dimforge/parry/pull/255/files#r1740956075

crates/parry3d/benches/all.rs
```diff
@@ -1,6 +1,8 @@
 #![feature(test)]
```
Vrixyz, line 1: Due to this feature, we can only use nightly. I'm not sure if this is avoidable.

crates/parry3d/Cargo.toml
```diff
@@ -89,6 +89,11 @@ oorandom = "11"
 ptree = "0.4.0"
 rand = { version = "0.8" }
 macroquad = "0.4.12"
+nalgebra = { version = "0.33", default-features = false, features = [
+    "libm",
```
sebcrozet, line 93: Is the `libm` feature really necessary for the benchmarks?

crates/parry3d/benches/all.rs
sebcrozet: Yeah, not sure either. Running benchmarks in nightly is fine.

crates/parry3d/benches/all.rs
```diff
@@ -10,3 +12,8 @@ mod bounding_volume;
 mod common;
 mod query;
 mod support_map;
+
+#[cfg(feature = "dim2")]
+type ConvexHull = Vec<Point<Real>>;
+#[cfg(feature = "dim3")]
+type ConvexHull = (Vec<Point<Real>>, Vec<[u32; 3]>);
```
sebcrozet, line 19:
```suggestion
```

crates/parry3d/benches/bounding_volume/mod.rs
sebcrozet: The `ConvexHull` shpe type has been renamed `ConvexPolyhedron` in 3D and `ConvexPolygon` in 2D.

crates/parry3d/benches/common/macros.rs
```diff
@@ -72,7 +72,7 @@ macro_rules! bench_method_gen (
 
                 unsafe {
                     let val: $tres = test::black_box($arg.get_unchecked(i).$method($(unref($args.get_unchecked(i)),)*));
-                    drop(val);
+                    let _ = val;
```
sebcrozet, line 75: Why was that change needed?

sebcrozet (commented):
Thank you for this PR! Note that the `ConvexHull` shpe type has been renamed `ConvexPolyhedron` in 3D and `ConvexPolygon` in 2D.

crates/parry3d/benches/common/macros.rs
Vrixyz: using `drop` results in a warning:

```rs
warning: calls to `std::mem::drop` with a value that implements `Copy` does nothing
   --> crates/parry3d/benches/bounding_volume/../common/macros.rs:75:21
```

`Aabb` and `BoundingSphere` are concerned. I don´t think this is an opportunity to make benchmarks more accurate (if we're unexpectedly copying).

Actually I'm not exactly sure why this drop is necessary. As we're in an unsafe call, I guess it might be useful, but not really used in current usage. It doesn´t hurt being here I guess.

crates/parry3d/Cargo.toml
Vrixyz: it's not, I'm removing it.

PR #258, opened 2024-09-03
Fix minor bug in mesh mesh intersection

Makogan:
One of the intersection bugs happens due to a mistake when setting up isometries.


Makogan:
@sebcrozet Fixed the bug with isometries not being applied correctly.

sebcrozet (approved):
Thanks!

PR #262, opened 2024-09-04
Propagate errors from mesh creation

Vrixyz:
- Follow up to https://github.com/dimforge/rapier/issues/717

This will have impacts on user facing API.

Also, an interesting observation on current error handling withing intersection testing, where parry has been leaking a private type through function signatures, which results in "unknown types" (untypable variables, and no links to source definition in rustdoc. small repro: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a414ede63130effa3b6e983d11a08bab

### DIFF
--- a/src/shape/trimesh.rs
+++ b/src/shape/trimesh.rs
@@ -846,6 +846,17 @@ impl TriMesh {
             )
         })
     }
+
+    #[cfg(feature = "dim3")]
+    /// Gets the normal of the triangle represented by `feature`.
+    pub fn feature_normal(&self, feature: FeatureId) -> Option<Unit<Vector<Real>>> {
+        match feature {
+            FeatureId::Face(i) => self
+                .triangle(i % self.num_triangles() as u32)
+                .feature_normal(FeatureId::Face(0)),
+            _ => None,
+        }
+    }
 }
 
 impl TriMesh {
--- a/src/shape/trimesh.rs
+++ b/src/shape/trimesh.rs
@@ -28,7 +28,7 @@ pub enum TopologyError {
     #[error("the triangle {0} has at least two identical vertices.")]
     BadTriangle(u32),
     /// At least two adjacent triangles have opposite orientations.
-    #[error("the triangles {triangle1} and {triangle2} sharing the edge {:?} have opposite orientations.", edge)]
+    #[error("the triangles {triangle1} and {triangle2} sharing the edge {edge:?} have opposite orientations.")]
     BadAdjacentTrianglesOrientation {
         /// The first triangle, with an orientation opposite to the second triangle.
         triangle1: u32,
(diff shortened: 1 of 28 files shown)


src/shape/trimesh.rs
```diff
@@ -54,6 +54,28 @@ impl fmt::Display for TopologyError {
...
 impl std::error::Error for TopologyError {}
 
+/// Indicated an inconsistency while building a triangle mesh.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum TriMeshBuilderError {
+    /// A triangle mesh must contain at least one triangle.
+    EmptyIndices,
+    /// Indicated an inconsistency in the topology of a triangle mesh.
+    TopologyError(TopologyError),
+}
```
Vrixyz, line 64: use thiserror ; see convexhullerror

src/shape/trimesh.rs
```diff
@@ -22,11 +22,13 @@ use crate::query::details::NormalConstraints;
...
 
 /// Indicated an inconsistency in the topology of a triangle mesh.
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[derive(thiserror::Error, Copy, Clone, Debug, PartialEq, Eq)]
 pub enum TopologyError {
     /// Found a triangle with two or three identical vertices.
+    #[error("the triangle {0} has at least two identical vertices.")]
     BadTriangle(u32),
     /// At least two adjacent triangles have opposite orientations.
+    #[error("the triangles {triangle1} and {triangle2} sharing the edge {:?} have opposite orientations.", edge)]
```
sebcrozet, line 31: I think this can be done now:
```suggestion
    #[error("the triangles {triangle1} and {triangle2} sharing the edge {edge:?} have opposite orientations.")]
```

sebcrozet (approved):
Thanks!

src/shape/trimesh.rs
```diff
@@ -415,15 +413,15 @@ impl TriMesh {
...
+        *self = TriMesh::with_flags(vertices, indices, self.flags).unwrap();
     }
 
     /// Create a `TriMesh` from a set of points assumed to describe a counter-clockwise non-convex polygon.
     ///
     /// This operation may fail if the input polygon is invalid, e.g. it is non-simple or has zero surface area.
     #[cfg(feature = "dim2")]
     pub fn from_polygon(vertices: Vec<Point<Real>>) -> Option<Self> {
-        triangulate_ear_clipping(&vertices).map(|indices| Self::new(vertices, indices))
+        triangulate_ear_clipping(&vertices).map(|indices| Self::new(vertices, indices).unwrap())
```
Vrixyz, line 424: this is incorrect and should propagate the error

Issue #253, opened 2024-08-13
epa3.rs panic in `closest_points`


bellwether-softworks:
Didn't want to hijack #246 since my conditions are different, but I am running into my own crash for the epa3.rs `*self.heap.peek()` statement.  I can reliably encounter it in the wild, but despite my best efforts to reproduce it outside of that environment I've been unable to.

I've modified the affected statement (and added some `Debug` annotations where appropriate) to be able to capture the state of the process at the time of failure:

```rust
let Some(mut best_face_id) = self.heap.peek().cloned() else {
    eprintln!("dim {}, vertices = {:?}, faces = {:?}, silhouette = {:?}, heap = {:?}", simplex.dimension(), self.vertices, self.faces, self.silhouette, self.heap);
    panic!("Failed to identify best face ID")
};
```

...which results in the following output:
```
[...]

Issue #246, opened 2024-07-29
Crash in epa3.rs 


ethereumdegen:
Getting a crash on this linein epa3 around line 313  i believe: 

 let mut best_face_id = *self.heap.peek().unwrap();
 
 I dont know what is causing it in my instance yet but why is this an unwrap?  seems unsafe.  
 
 ```
 thread 'Compute Task Pool (2)' panicked at /home/andy/.cargo/registry/src/index.crates.io-6f17d22bba15001f/parry3d-0.15.1/src/query/epa/epa3.rs:313:50:
called `Option::unwrap()` on a `None` value

```

ethereumdegen:
I have narrowed it down more.. Spawning a Collider with a Sensor but without a RigidBody or Dynamicbody  reliably causes this crash

wlinna:
@sebcrozet I think this issue can be closed now. Two PRs have been merged to address this already :)

PR #263, opened 2024-09-05
Prevent crash in epa2 and epa3 ::closest_points if all is_project

wlinna:
EDIT: I implemented a fix for epa3 that should also make the test case work as expected

Alleviates/fixes these issues:
https://github.com/dimforge/parry/issues/253
https://github.com/dimforge/parry/issues/246

Related Discord thread:
https://discord.com/channels/507548572338880513/1274773766811160646

`EPA::closest_point ` crashes at line `let mut best_face_id = *self.heap.peek().unwrap();` because `heap` is empty. This happens when we reach the second branch of `closest_point` and all of the `proj_inside` variables are false.

The measure taken in this PR is to log the problem and return `None`. While the result might be incorrect, it avoids the panic.
This is only a half-measure and someone who understands EPA and/or this implementation of it should write a proper fix.

The included unit test is programmed to fail because in reality there should be an intersection.


src/query/epa/epa2.rs
```diff
@@ -226,28 +226,36 @@ impl EPA {
...
+            if proj_inside3 {
                 let dist3 = self.faces[2].normal.dot(&self.vertices[2].point.coords);
                 self.heap.push(FaceId::new(2, -dist3)?);
             }
+
+            if !(proj_inside1 || proj_inside2 || proj_inside3) {
+                // Related issues:
+                // https://github.com/dimforge/parry/issues/253
+                // https://github.com/dimforge/parry/issues/246
+                log::debug!("Hit unexpected state in EPA: proj_inside1, proj_inside2 and proj_inside3 are all false. At least one of them should be true.");
```
sebcrozet, line 256:
```suggestion
                log::debug!("Hit unexpected state in EPA: failed to project the origin on the initial simplex.");
```

src/query/epa/epa3.rs
```diff
@@ -284,6 +284,14 @@ impl EPA {
                 let dist4 = self.faces[3].normal.dot(&self.vertices[3].point.coords);
                 self.heap.push(FaceId::new(3, -dist4)?);
             }
+
+            if !(proj_inside1 || proj_inside2 || proj_inside3 || proj_inside4) {
+                // Related issues:
+                // https://github.com/dimforge/parry/issues/253
+                // https://github.com/dimforge/parry/issues/246
+                log::debug!("Hit unexpected state in EPA: proj_inside1, proj_inside2, proj_inside3 and proj_inside4 are all false. At least one of them should be true.");
```
sebcrozet, line 292:
```suggestion
                log::debug!("Hit unexpected state in EPA: failed to project the origin on the initial simplex.");
```

sebcrozet (approved):
Good catch, thanks!

src/query/epa/epa3.rs
wlinna: I just realized something. Does the projection really "fail"? In my test, all of the `Face::new` calls report `OnEdge`  location for the projection instead of `OnFace`. But could we not also consider `OnVertex` and `OnEdge` workable locations for the projections?

wlinna:
```rust
);
        let (_, loc) = tri.project_local_point_and_get_location(&Point::<Real>::origin(), true);

        match loc {
            TrianglePointLocation::OnFace(_, bcoords) => {
                (Self::new_with_proj(vertices, bcoords, pts, adj), true)
            }
            _ => (Self::new_with_proj(vertices, [0.0; 3], pts, adj), false),
        }
```

Was there an assumption, that if the location is `OnEdge` or `OnVertex`, origin was not touching the triangle in the first place? In this case the origin seems to be touching an edge already, so maybe instead of relying on `OnFace`, we should check whether the projection is sufficiently close to the origin? Two of the projections are so close to origin, that the distance could be attributed to numerical instability:

```
[crates/parry3d/../../src/query/epa/epa3.rs:101:9] &proj = PointProjection {
    is_inside: true,
    point: [
        0.0,
        0.0,
        -1.1920929e-7,
    ],
}
[crates/parry3d/../../src/query/epa/epa3.rs:101:9] &proj = PointProjection {
    is_inside: true,
    point: [
        0.0,
        0.0,
        -1.1920929e-7,
    ],
}
```

wlinna:
If I change the body of `Face::new` to this, all the tests, including the new one, pass

```
 pub fn new(vertices: &[CSOPoint], pts: [usize; 3], adj: [usize; 3]) -> (Self, bool) {
        let tri = Triangle::new(
            vertices[pts[0]].point,
            vertices[pts[1]].point,
            vertices[pts[2]].point,
        );

        let (proj, loc) = tri.project_local_point_and_get_location(&Point::<Real>::origin(), true);

        match loc {            
            TrianglePointLocation::OnVertex(_) | TrianglePointLocation::OnEdge(_, _) => {
                let eps = crate::math::DEFAULT_EPSILON;
                let eps_tol = eps * 100.0;
                let is_inside = proj.is_inside_eps(&Point::<Real>::origin(), eps_tol);
                (Self::new_with_proj(vertices, loc.barycentric_coordinates().unwrap(), pts, adj), is_inside)
            },            
            TrianglePointLocation::OnFace(_, bcoords) => {
                (Self::new_with_proj(vertices, bcoords, pts, adj), true)
            }
            _ => (Self::new_with_proj(vertices, [0.0; 3], pts, adj), false),
        }
    }
```

Do you think this could work?

wlinna:
I went ahead and pushed this change. We can always revert the commit if it doesn't make sense.
Please review the change :smiley:

wlinna:
Sorry, this PR has turned into a mess because I did a merge commit.

wlinna:
@sebcrozet Hello, I wonder if you could review my PR and merge it? I don't want to pressure you, but I've noticed that you recently merged https://github.com/dimforge/parry/pull/282 which only partially addresses the fallibility issue, while this one should fix the underlying issue. Also, unlike the recently merged PR, this one addresses epa2 as well.

Would you please consider reviewing this?

sebcrozet:
This looks good, thank you! And sorry for the delay.

PR #265, opened 2024-09-10
fix point_in_poly2d for self-intersecting polygons

sebcrozet:
In some cases where the polygon loops on itself, the winding number was incorrect.


sebcrozet:
Closing as it is already in`master`.

PR #268, opened 2024-09-12
Fix wavefront tests + new test for different pos + add them in ci

Vrixyz:
- Follow up to https://github.com/dimforge/parry/pull/258

### DIFF
--- a/crates/parry3d-f64/Cargo.toml
+++ b/crates/parry3d-f64/Cargo.toml
@@ -44,6 +44,7 @@ simd-stable = ["simba/wide", "simd-is-enabled"]
 simd-nightly = ["simba/portable_simd", "simd-is-enabled"]
 enhanced-determinism = ["simba/libm_force", "indexmap"]
 parallel = ["rayon"]
+# Adds `TriMesh:to_obj_file` function.
 wavefront = ["obj"]
 alloc = []
 improved_fixed_point_support = []
@@ -88,3 +89,7 @@ thiserror = { version = "1", optional = true }
 oorandom = "11"
 ptree = "0.4.0"
 rand = { version = "0.8" }
+
+[package.metadata.docs.rs]
+rustdoc-args = ["-Zunstable-options", "--generate-link-to-definition"]
+features = ["wavefront"]
--- a/src/transformation/mesh_intersection/mesh_intersection.rs
+++ b/src/transformation/mesh_intersection/mesh_intersection.rs
@@ -5,8 +5,6 @@ use crate::query::{visitors::BoundingVolumeIntersectionsSimultaneousVisitor, Poi
 use crate::shape::{TriMesh, Triangle};
 use crate::utils;
 use na::{Point3, Vector3};
-#[cfg(feature = "wavefront")]
-use obj::{Group, IndexTuple, ObjData, Object, SimplePolygon};
 use rstar::RTree;
 use spade::{ConstrainedDelaunayTriangulation, InsertionError, Triangulation as _};
 use std::collections::BTreeMap;
@@ -668,9 +666,10 @@ fn merge_triangle_sets(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::shape::TriMeshFlags;
-    use crate::transformation::wavefront::*;
+    use crate::bounding_volume::{bounding_sphere, BoundingSphere};
+    use crate::shape::{Ball, Cuboid, TriMeshFlags};
     use obj::Obj;
+    use obj::ObjData;
 
     #[test]
     fn test_same_mesh_intersection() {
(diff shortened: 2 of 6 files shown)


src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -708,6 +707,33 @@ mod tests {
...
+        sphere_mesh.set_flags(TriMeshFlags::all()).unwrap();
+        let mut cuboid_mesh = TriMesh::new(cuboid.0, cuboid.1);
+        cuboid_mesh.set_flags(TriMeshFlags::all()).unwrap();
+
+        let res = intersect_meshes(
+            &Isometry::translation(1.0, 0.0, 0.0),
+            &cuboid_mesh,
+            false,
+            &Isometry::translation(2.0, 0.0, 0.0),
+            &sphere_mesh, //.clone().scaled(&Vector3::new(1.001, 1.001, 1.001)),
```
Vrixyz, line 724: That will have to be removed, but I added that because setting both positions to `200;0;0` resulted in a lot of holes in the resulting intersection. I can imagine this is floating point imprecision but it was surprising.

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -708,6 +707,33 @@ mod tests {
...
+            false,
+        )
+        .unwrap()
+        .unwrap();
+
+        let _ = res.to_obj_file(&PathBuf::from("test_non_origin_pos1_pos2_intersection.obj"));
+
+        let bounding_sphere = res.local_bounding_sphere();
+        assert!(bounding_sphere.center == Point3::new(1.5, 0.0, 0.0));
+        assert_relative_eq!(2.0615528, bounding_sphere.radius, epsilon = 1.0e-5);
```
Vrixyz, line 734: I'm only testing bounding sphere, because testing the whole trimesh is not easy, as it doesn´t seem deterministic (I had different results between runs)

src/transformation/mesh_intersection/mesh_intersection.rs
sebcrozet: Interesting. I think it’s worth opening an issue with the example that generates holes.

sebcrozet (approved):
Thanks!

src/transformation/mesh_intersection/mesh_intersection.rs
Vrixyz: I opened https://github.com/dimforge/parry/issues/270

crates/parry3d-f64/Cargo.toml
```diff
@@ -44,6 +44,7 @@ simd-stable = ["simba/wide", "simd-is-enabled"]
 simd-nightly = ["simba/portable_simd", "simd-is-enabled"]
 enhanced-determinism = ["simba/libm_force", "indexmap"]
 parallel = ["rayon"]
+# Adds `TriMesh:to_obj_file` function.
 wavefront = ["obj"]
```
Vrixyz, line 48: Adding documentation on `cargo.toml` is idiomatic to rust ecosystem (see serde or bevy)

crates/parry3d-f64/Cargo.toml
```diff
@@ -88,3 +89,7 @@ thiserror = { version = "1", optional = true }
 oorandom = "11"
 ptree = "0.4.0"
 rand = { version = "0.8" }
+
+[package.metadata.docs.rs]
+rustdoc-args = ["-Zunstable-options", "--generate-link-to-definition"]
+features = ["wavefront"]
```
Vrixyz, line 95: This allows to see the function in rustdoc.rs

crates/parry3d-f64/Cargo.toml
```diff
@@ -88,3 +89,7 @@ thiserror = { version = "1", optional = true }
 oorandom = "11"
 ptree = "0.4.0"
 rand = { version = "0.8" }
+
+[package.metadata.docs.rs]
+rustdoc-args = ["-Zunstable-options", "--generate-link-to-definition"]
```
Vrixyz, line 94: this is already present in parry3d (we're in parry3d-f64 here)

src/transformation/mod.rs
```diff
@@ -40,4 +40,4 @@ mod to_trimesh;
 pub mod utils;
 
 #[cfg(feature = "wavefront")]
-pub mod wavefront;
+mod wavefront;
```
Vrixyz, line 43: This module doesn´t have to be public, as it's implementing functions directly on `TriMesh`, which is public.

Making it public warns about undocumented module, as well as having nothing to show in it (the added function is within `TriMesh`)

Issue #277, opened 2024-10-06
Give option to not remove vertices when constructing ConvexPolyline and other such shapes


Ughuuu:
Right now if I create a `ConvexPolygon` I use `from_convex_polyline`, however sometimes this will remove some points. Give option to not remove the points (eg. default constructor or simpler constructor).

WaffeFIN:
I too have this issue. I would suggest a change to the epsilon used in _from_convex_polyline_ making it much smaller:
```rust
let eps = 2.0 * crate::math::DEFAULT_EPSILON;
```

It would be nice to have the tolerance/epsilon as a parameter, in which case it could be disabled by making it zero. I bet that change would be quite disruptive though.

Related to this matter, here are also some unit tests I've added to my local version of parry2d (convex_polygon.rs):
```rust
#[test]
fn test_convex_polyline_for_stretched_quad() {
    let vertices = vec![
        Point::new(-35.0, 0.0),
        Point::new(38.0, 0.0),
        Point::new(39.0, 1.0),
        Point::new(17.0, 1.0), 
    ];
    let expected_points = [
        Point::new(-35.0, 0.0),
        Point::new(38.0, 0.0),
        Point::new(39.0, 1.0),
        Point::new(17.0, 1.0), 
    ]; // No change

    let collider = ConvexPolygon::from_convex_polyline(vertices).unwrap();

    assert_eq!(collider.points(), expected_points);
}

#[test]
fn test_convex_polyline_for_basicly_triangle() {
    let vertices = vec![
        Point::new(-1000.0, 0.0),
        Point::new(1000.0, 0.0),
        Point::new(1000.0, 1.0),
        Point::new(-0.01, 0.5), 
    ];
    let expected_points = [
        Point::new(-1000.0, 0.0),
        Point::new(1000.0, 0.0),
        Point::new(1000.0, 1.0),
    ];

    let collider = ConvexPolygon::from_convex_polyline(vertices).unwrap();

    assert_eq!(collider.points(), expected_points);
}
```

PR #278, opened 2024-10-06
Create constructor for ConvexPolyline that does not remove any vertices

Ughuuu:
- Fixes https://github.com/dimforge/parry/issues/277
In Godot Rapier, Godot sends vertices that it uses to draw the polygon. If Parry removes some vertices, then the physics will not work correct. We need the extra vertices even if collinear, and ideally have the collinearity check done in Godot at some point or sent as a warning to console. For now a function that accepts points and doesn't remove any is needed for this use case.


sebcrozet:
Thank you! I just renamed the new methods to `from_convex_polyline_unmodified`.

PR #284, opened 2024-11-06
Degenerate face normal enum

ThierryBerger:


src/query/contact/contact_support_map_support_map.rs
```diff
@@ -19,12 +19,16 @@ where
...
+        GJKResult::ClosestPoints(point1, point2_1, Ok(normal1)) => {
             let dist = (point2_1 - point1).dot(&normal1);
             let point2 = pos12.inverse_transform_point(&point2_1);
             let normal2 = pos12.inverse_transform_unit_vector(&-normal1);
             Some(Contact::new(point1, point2, normal1, normal2, dist))
         }
+        GJKResult::ClosestPoints(_, _, Err(_)) => {
+            // TODO: propagate the error.
+            None
+        }
```
Vrixyz, line 31: this is a breaking change, which would end up in breaking access succeeding before (but probably behaving slightly incorrectly).

Vrixyz: I consider support map without normal information incorrect

Vrixyz: TODO: add warn! + test example 3d convex decomposition

PR #286, opened 2024-11-12
Improve robustness of mesh/mesh intersection

sebcrozet:
This PR further improves the robustness the mesh/mesh intersection for a better handling of pars of meshes with coplanar triangles. This was motivated by the need to fix a case where the same mesh would be cut multiple times by the exact same second mesh. All cuts except for the first one result in degenerate cases where many triangles are coplanar.

This applies the following changes:
- Fix an issue where the orientation of a triangle output by the convex-polygons intersection algorithm, when given one triangle fully inside the other, is incorrect.
- Fix an issue where coplanar triangles could be output twice by the mesh/mesh intersection code.
- Improve robustness of convex-polygons intersection by avoiding the incorrect in-out state to be set when traversing an ambiguous configuration (when a point of one polygon lies exactly on an edge of the other polygon).
- Fix an issue where the vertices of both polygons might be output by the convex-polygons intersection algorithm if one triangle is fully inside the other but has its vertices exactly on the edges of the other polygon.

Tolerances can now be specified for the mesh/mesh intersection (the method has been made public), and for the convex-polygons intersection.

The variable names in convex-polygons intersection have be changed for better readability.
Note that this introduces a O(n²) complexity for the case where a polygon is fully inside the other one. This can be addressed in a future PR.

### DIFF
--- a/src/transformation/mesh_intersection/mesh_intersection.rs
+++ b/src/transformation/mesh_intersection/mesh_intersection.rs
@@ -21,13 +21,14 @@ struct HashableTriangleIndices([u32; 3]);
 
 impl From<[u32; 3]> for HashableTriangleIndices {
     fn from([a, b, c]: [u32; 3]) -> Self {
-        let (sa, sb, sc) = crate::utils::sort3(&a, &b, &c);
+        let (sa, sb, sc) = utils::sort3(&a, &b, &c);
         HashableTriangleIndices([*sa, *sb, *sc])
     }
 }
 
 /// Metadata that specifies thresholds to use when making construction choices
 /// in mesh intersections.
+#[derive(Copy, Clone, PartialEq, Debug)]
 pub struct MeshIntersectionTolerances {
     /// The smallest angle (in radians) that will be tolerated. A triangle with
     /// a smaller angle is considered degenerate and will be deleted.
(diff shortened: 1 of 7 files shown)


src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -143,33 +163,40 @@ pub fn intersect_meshes_with_tolerances(
     // 4: Initialize a new mesh by inserting points into a set. Duplicate points should
     // hash to the same index.
     let mut point_set = RTree::<TreePoint, _>::new();
-    let mut topology_indices = Vec::new();
+    let mut topology_indices = HashMap::new();
```
Makogan, line 166: I am a bit scared of this. The non-determinism might create hard to reproduce bugs in the future, since the traversal order of the faces will change between invocations. I would like to suggest a BTreeMap. It won't necessarily be slower in most cases because the hashmap iteration will suffer from more cache misses.

src/transformation/mesh_intersection/mesh_intersection.rs
```diff
@@ -243,7 +271,10 @@ pub fn intersect_meshes_with_tolerances(
     let vertices: Vec<_> = vertices.iter().map(|p| Point3::from(p.point)).collect();
 
     if !topology_indices.is_empty() {
-        Ok(Some(TriMesh::new(vertices, topology_indices)?))
+        Ok(Some(TriMesh::new(
+            vertices,
+            topology_indices.into_values().collect(),
```
Makogan, line 276: Very scared of the non-determinism of the hashmap.

Makogan (approved):

src/transformation/mesh_intersection/mesh_intersection.rs
Makogan: Cosnidering we even tolerated an n^2 algorithm elsewhere. I wold want to suggest switching to a BTreeMap, or at least benchmark to ensure that the performance gains justify a hashmap.

src/transformation/mesh_intersection/mesh_intersection.rs
sebcrozet: Agreed, I’ll switch to `crate::utils::hashmap::HashMap` which is deterministic.

PR #273, opened 2024-09-20
Address a few TODOs

Vrixyz:

PR #293, opened 2024-12-05
IntersectionCompositeShapeShapeVisitor is now able to return partId

Vrixyz:
- Follow up to https://github.com/dimforge/parry/pull/273#issuecomment-2496130758 ; Thanks @wlinna!

### DIFF
--- a/src/query/intersection_test/intersection_test_composite_shape_shape.rs
+++ b/src/query/intersection_test/intersection_test_composite_shape_shape.rs
@@ -1,9 +1,12 @@
 use crate::bounding_volume::SimdAabb;
-use crate::math::{Isometry, Real, SIMD_WIDTH};
-use crate::partitioning::{SimdVisitStatus, SimdVisitor};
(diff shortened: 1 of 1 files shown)


src/query/intersection_test/intersection_test_composite_shape_shape.rs
```diff
@@ -114,3 +123,55 @@ where
         SimdVisitStatus::MaybeContinue(mask)
     }
 }
+
+impl<'a, D, G1> SimdBestFirstVisitor<G1::PartId, SimdAabb>
+    for IntersectionCompositeShapeShapeVisitor<'a, D, G1>
```
Vrixyz, line 128: Should this be kept ?

sebcrozet: Yes

sebcrozet (changes requested):
Looking good, thanks! Just need to remove the best first impl.

sebcrozet (approved):

PR #295, opened 2024-12-06
Fix infinite loop in intersection_with_local_plane

nikita240:
I added a test case `trimesh_plane_square_intersection`. If you run this test with the current master, it will get stuck in an infinite loop. We're missing an exit condition.


ThierryBerger (approved):

sebcrozet:
Thank you. Sorry for getting to this late. This seems to already be fixed on `master` now as I cannot reproduce it. Please reopen if that’s not the case.

PR #302, opened 2025-01-10
Fix `clippy::question_mark` lint

waywardmonkeys:


waywardmonkeys:
@Vrixyz has since done the same change. (But not the ones in my other PRs.)

PR #306, opened 2025-01-10
Update `thiserror` to 2.x

waywardmonkeys:
This brings `no_std` support, so this is now able to be used in a `no_std` context, so always build it, but enable `std` support when building parry for `std`.


Vrixyz:
Thanks!

PR #307, opened 2025-01-13
Custom dispatcher example

ThierryBerger:
- In order to improve understanding around issues like https://github.com/dimforge/rapier/issues/777 ; an example showcasing a custom dispatcher might be useful.

I'd like something more advanced than a forward-to-default dispatcher though :thinking: .

### DIFF
--- a/crates/parry3d/examples/custom_dispatcher.rs
+++ b/crates/parry3d/examples/custom_dispatcher.rs
@@ -12,6 +12,8 @@ use parry3d::{
     shape::{Ball, Cuboid, FeatureId, Shape, ShapeType, TypedShape},
 };
 
+pub struct CustomBall(pub Ball);
+
 fn main() {
     let cube = Cuboid::new(Vector::new(1.0, 1.0, 1.0));
     let ball = CustomBall(Ball::new(1.0));
@@ -23,8 +25,6 @@ fn main() {
 
     dbg!(contact);
 }
-
-pub struct CustomBall(pub Ball);
 
 impl PointQuery for CustomBall {
     fn project_local_point(&self, pt: &Point<f32>, solid: bool) -> PointProjection {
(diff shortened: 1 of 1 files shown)


crates/parry3d/examples/custom_dispatcher.rs
```diff
@@ -0,0 +1,167 @@
...
+
+impl QueryDispatcher for CustomBallDispatcher {
+    fn intersection_test(
+        &self,
+        pos12: &Isometry<f32>,
+        g1: &dyn Shape,
+        g2: &dyn Shape,
+    ) -> Result<bool, Unsupported> {
+        let dispatcher = DefaultQueryDispatcher;
+        dispatcher.intersection_test(pos12, g1, g2)
```
Vrixyz, line 99: call directly `intersection_test_ball_ball` to show custom dispatch ; only then fallbackto default's

Vrixyz: + test if that workds :sweat_smile:

crates/parry3d/examples/custom_dispatcher.rs
```diff
@@ -0,0 +1,327 @@
...
+                ));
+            }
+            (Some(ball1), None) => {
+                let dispatcher = DefaultQueryDispatcher;
+                dispatcher.distance(pos12, &ball1.0, g2)
+            }
+            (None, Some(ball2)) => {
+                let dispatcher = DefaultQueryDispatcher;
+                dispatcher.distance(pos12, g1, &ball2.0)
+            }
```
Vrixyz, line 150: These branches are not necessary for the example, we should replace them by a comment saying "implement the algorithm for each shape type combination you want to support."

crates/parry3d/examples/custom_dispatcher.rs
```diff
@@ -0,0 +1,269 @@
...
+};
+
+pub struct CustomShape(pub Ball);
+
+fn main() {
+    let cube = Cuboid::new(Vector::new(1.0, 1.0, 1.0));
+    let custom_shape = CustomShape(Ball::new(1.0));
+
+    let pos12 = Isometry::identity();
+    let dispatcher = CustomShapeDispatcher;
```
Vrixyz, line 22: consider using [QueryDispatcher::chain](https://github.com/dimforge/parry/blob/8b6f659b1334eecd534f199d5ccacf366c3e807e/src/query/query_dispatcher.rs#L116) ?

PR #312, opened 2025-02-01
Fix shape-heightfield shape cast

unpairedbracket:
This was causing shapecasts against heightfields to fall through to Unsupported, if and only if the heightfield was in the second position


Vrixyz (approved):
thanks!

PR #314, opened 2025-02-06
Add mutable accessor for a VoxelSet's voxels.

dbenson24:
This is to allow external code to call into VHACD with a set of voxels that it computes without forcing it to generate a bunch of geometry to then get revoxelized.

### DIFF
--- a/src/transformation/voxelization/voxel_set.rs
+++ b/src/transformation/voxelization/voxel_set.rs
@@ -165,6 +165,11 @@ impl VoxelSet {
         &self.voxels
     }
 
+    /// A mutable reference to the set of voxels.
+    pub fn voxels_mut(&mut self) -> &mut Vec<Voxel> {
+        &mut self.voxels
+    }
+
     /// Update the bounding box of this voxel set.
     pub fn compute_bb(&mut self) {
         let num_voxels = self.voxels.len();


dbenson24:
I would also be interested in adding the following helper to SharedShape if it looks reasonable to you
```rs
impl SharedShape {
 pub fn voxel_convex_decomposition(voxels: VoxelSet, config: &VHACDConfig) -> SharedShape {
        let decomp = VHACD::from_voxels(config, voxels);
        let mut parts = Vec::new();
        for vertices in decomp.compute_convex_hulls(0) {

            if let Some(convex) = SharedShape::convex_polyline(vertices) {
                parts.push((Isometry::identity(), convex));
            }
        }
        SharedShape::compound(parts)
}

```` 

That looks like

ThierryBerger (commented):
I think this is reasonable, but I'm a bit worried about how to keep "invariants" true, I believe an incorrect `min_bb_voxels`or `intersections` is asking for trouble, what's your stance on that?

dbenson24:
I think that just documenting this potential concern could be enough? This is a pretty low level API, and if you give VHACD an incorrect VoxelSet and get bad results it's on you. It'll also be up to the caller to keep the voxels unique and within the dimensions.

I opted to just expose the vec because it was the smallest change. I think another API that would probably work for most would be some way to populate the VoxelSet from an iterator. Maybe something like set_voxels(voxels: impl Iter<Item=Voxel>) which recalculates the min_bb right after 



For intersections that should always be empty since there is no original geometry to intersect with.

dbenson24:
The use case I'm specifically working on trying to improve is the current technique that people are using to generate colliders for sprites. Right now you have to try and convert the pixels into a set of lines to feed in to the voxelizer before getting the decomposition. If you just feed the pixels of the sprite into the decomp as voxels, you get to skip the performance cost of the voxelization and any inaccuracies introduced by the multiple conversions.

PR #321, opened 2025-03-05
fix trimesh inertia tensor

### LINKED ISSUE #320: Incorrect angular momentum tensor for trimesh
The angular momentum tensor computed for a trimesh has it's xy and xz components swapped. This can be fixed by swapping the locations of `b1` and `c1` in the final line of `mass_properties::details::tetrahedron_unit_inertia_tensor_wrt_point`.

krsnik02:
Correct the calculation of the trimesh inertia tensor. Fixes #320 .


Vrixyz:
Thanks!

- I'll take it from here + add a test + changelog over at https://github.com/dimforge/parry/pull/331 -> let's discuss there from now on (I'll include you in commit co-author on the fix).

PR #325, opened 2025-03-07
Derive `Copy` for `VHACDParameters`

waywardmonkeys:
This is to make something in Rapier be able to impl `Copy`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## Unreleased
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
There's a comment in Rapier that this will help address:

```
// TODO: implement Copy once we add a Copy implementation for VHACDParameters.
```

Vrixyz:
thanks! I'll add a changelog line and merge

PR #328, opened 2025-03-19
Fix most `clippy::doc_overindented_list_items` lints

waywardmonkeys:
The remaining ones are more opinionated.


Vrixyz:
Thanks!

Out of curiosity, how do these improve end user experience ? I believe these are not reported when brought by dependencies right ? I'm not sure

Should we add more clippy lints to the CI to help avoiding future regression?

waywardmonkeys:
They are new in nightly clippy, and will make their way to stable. They do detect other cases where indentation is actually wrong, so making sure it is consistent will help.

Vrixyz:
Thanks :heart:  !

PR #331, opened 2025-03-27
Fix trimesh inertia tensor

ThierryBerger:
- Coming from #321

I added a test case to confirm and avoid further regressions.


Vrixyz:
44acba15a699a4925f6cd618f73fe795691f45cd doesn't include the fix, which shows the current behaviour is incorrect.

PR #334, opened 2025-04-03
Fix wrong sign in `MassProperties::construct_shifted_inertia_matrix`

Jondolf:
# Objective

Parry's formula for the 3D shifted angular inertia tensor is wrong, causing compound shapes with offsets to have an incorrect angular inertia. This results in wonky behavior in physics engines such as Rapier and Avian.

An example of this: A 1x3x1 cuboid collider in bevy_rapier3d has a principal angular inertia of `[2.5, 0.5, 2.5]`. However, a compound collider of three 1x1x1 cubes stacked on top of each other has a principal angular inertia of `[2.5, 4.5, 2.5]`.

Logs

Spawning dynamic bodies for the two colliders with a slight rotation about the x axis and some initial angular velocity, this results in the following behavior:

https://github.com/user-attachments/assets/0bcc9cc4-10ed-467b-9907-a85b6092f5f8

The 1x3x1 cuboid spins a bit at the start, falls on its side, and goes to rest quickly, like expected. However, the compound shape has a lot more angular inertia about the y axis, causing it to spin a lot longer and even start rolling on the ground in a bizarre way.

## Solution

`MassProperties::construct_shifted_inertia_matrix` uses the [parallel axis theorem](https://en.wikipedia.org/wiki/Parallel_axis_theorem) to compute the moment of inertia at a given offset. More specifically, it uses the [generalization for tensors](https://en.wikipedia.org/wiki/Parallel_axis_theorem#Tensor_generalization).

The correct formula looks like the following:

```
I_shifted = I + m * (dot(R, R) E_3x3 - outer_product(R, R))
```

where `E_3x3` is the 3x3 identity matrix and `R` is the displacement vector (i.e. shift). However, Parry incorrectly *adds* the outer product of the displacement rather than subtracting it.

Fixing this results in the correct values and much more reasonable behavior.

Logs

https://github.com/user-attachments/assets/8b73fb08-f440-41b9-8eb3-2d2bf388524c

### DIFF
--- a/src/mass_properties/mass_properties.rs
+++ b/src/mass_properties/mass_properties.rs
@@ -592,26 +592,35 @@ mod test {
 
     #[test]
     #[cfg(feature = "alloc")]
-    fn mass_properties_sum_shifted() {
+    fn mass_properties_compound() {
         use na::Isometry;
 
-        use crate::{math::Vector, shape::Cuboid};
+        use crate::{
+            math::Vector,
+            shape::{Compound, Cuboid, SharedShape},
+        };
 
         // Compute the mass properties of a compound shape made of three 1x1x1 cuboids.
-        let mp = Cuboid::new(Vector::repeat(0.5)).mass_properties(1.0);
-        let sum = [
-            mp,
-            mp.transform_by(&Isometry::from_parts(
-                Vector::y().into(),
-                Default::default(),
-            )),
-            mp.transform_by(&Isometry::from_parts(
-                (-Vector::y()).into(),
-                Default::default(),
-            )),
-        ]
-        .into_iter()
-        .sum::<MassProperties>();
+        let shape = Cuboid::new(Vector::repeat(0.5));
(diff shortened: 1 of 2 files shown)


Vrixyz (approved):
Thanks! I appreciate the PR description, changelog and tests alongside the bugfix <3

src/mass_properties/mass_properties.rs
```diff
@@ -590,10 +590,49 @@ mod test {
...
+            mp.transform_by(&Isometry::from_parts(
+                (-Vector::y()).into(),
+                Default::default(),
+            )),
+        ]
+        .into_iter()
+        .sum::<MassProperties>();
+
+        // Check that the mass properties of the compound shape match the mass properties
+        // of a single cuboid 1x3x1
```
Jondolf, line 617:
```suggestion
        // Check that the mass properties of the compound shape match the mass properties
        // of a single 1x3x1 cuboid.
```

Issue #264, opened 2024-09-06
Should convex decomposition return an option or a result?


hakolao:
I ran into
```
A compound shape must contain at least one shape.
```

I'm generating lots of convex decomposition colliders like this:

```rust
        let contours = self.contours_with_holes_flat(bitmap_to_pixels, simplify_epsilon);
        if contours.len() < 4 {
            return vec![];
        }
        let indices = (0..contours.len() as u32 - 1)
            .map(|i| [i, i + 1])
            .collect::<Vec<[u32; 2]>>();
        let collider = ColliderBuilder::convex_decomposition_with_params(
            &contours,
            &indices,
            &VHACDParameters {
                resolution: 64,
                fill_mode,
                ..Default::default()
            },
        )
        .build();
```

The assertion in `parry2d-0.17.0\src\shape\compound.rs:35:9` causes a panic, and can't do much about it. I don't really know when exactly the function fails to generate a shape, and my attempt to not let too small contours only helps for some other panics...

I would prefer no panics, and let the user handle the issues (Option or Result).

PR #348, opened 2025-06-19
Fallible methods for decomposition

### LINKED ISSUE #347: Convex decomposition can panic
A path to panic can be reached through convex decomposition, which is used in a fair number of api calls. It usually 

- https://github.com/dimforge/parry/issues/65#issuecomment-2626499350
- https://github.com/Jondolf/avian/issues/645

A backtrace and exact reproduction would be appreciated, but I think the origin is from:
- https://github.com/Vrixyz/parry/blob/ba60bdc5efe865cfce846edd570f0a10c45701ab/src/transformation/convex_hull3/convex_hull.rs#L12-L13
- called from: https://github.com/Vrixyz/parry/blob/ba60bdc5efe865cfce846edd570f0a10c45701ab/src/transformation/voxelization/voxel_set.rs#L688

A path to resolution could be to use `try_convex_hull` and bubble the error up wherever possible.

ThierryBerger:
- improves #264 ; but the assert is still here.
- Fix #347 

This approach is to wrap computation results into `Result`s when it makes sense.

We can probably go further into this approach, and really hunt for any misconfiguration and return an error whenever the inputs don't make sense, but let's discuss this first.

I considered adding `try` variants, but it doesn't add too much value so I went straight to migrating to fallible everywhere.


src/shape/shared_shape.rs
```diff
@@ -319,7 +319,7 @@ impl SharedShape {
         vertices: &[Point<Real>],
         indices: &[[u32; DIM]],
         params: &VHACDParameters,
-    ) -> Self {
+    ) -> Option<Self> {
```
Vrixyz, line 322: This is a bit complicated to promote to fallible, because the input may contain multiple incorrect shapes.

behavior on master is to only return the shapes that the algorithm could make, and crash when NO shapes are able to be made.

This PR improves this by not crashing. It may be interesting to have more information about why it failed (and how to fix).

src/transformation/vhacd/vhacd.rs
```diff
@@ -36,6 +37,7 @@ pub(crate) struct CutPlane {
 }
 
 /// Approximate convex decomposition using the VHACD algorithm.
+#[derive(Debug)]
```
Vrixyz, line 40: I added a few debug derive to help me with degugging

src/transformation/vhacd/vhacd.rs
```diff
@@ -239,12 +240,18 @@ impl VHACD {
                     &mut right_ch_pts,
                     &mut left_ch_pts,
                 );
-                right_ch = convex_hull(&right_ch_pts);
-                left_ch = convex_hull(&left_ch_pts);
+                right_ch = convex_hull(&right_ch_pts)
+                    .unwrap_or_else(ConvexHullError::into_incorrect_empty);
```
Vrixyz, line 244: I'm not a fan of this behavior, but I think it makes sense to have an indicator at call site that the return may be incorrect (empty).

src/transformation/voxelization/voxelized_volume.rs
```diff
@@ -192,6 +193,7 @@ impl VoxelizedVolume {
         fill_mode: FillMode,
         keep_voxel_to_primitives_map: bool,
     ) -> Self {
+        debug_assert!(resolution > 1);
```
Vrixyz, line 196: I guess this should be a normal assert

src/shape/shared_shape.rs
Vrixyz: add a todo to improve error reporting

src/transformation/convex_hull_utils.rs
```diff
@@ -1,8 +1,58 @@
-use crate::math::Real;
+use alloc::vec::Vec;
```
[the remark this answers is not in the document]
Vrixyz: actually no, transformation is only here when alloc is enabled.

PR #355, opened 2025-06-28
Return error from `check_convex_hull`

### LINKED ISSUE #352: `convex_hull3::check_convex_hull` has panics
At the current moment, `convex_hull3::check_convex_hull` does not return any errors and panics internally that makes catching error complicating.
The panic is, for example, here:
https://github.com/dimforge/parry/blob/fd61591cf881436faa3e99385f4947b1b942d9fe/src/transformation/convex_hull3/validation.rs#L93

Routine source code:
https://github.com/dimforge/parry/blob/fd61591cf881436faa3e99385f4947b1b942d9fe/src/transformation/convex_hull3/validation.rs#L30-L99

It would be nice to provide ConvexHullError instead of panics.

foxtran:
Fixes #352


src/transformation/convex_hull3/validation.rs
```diff
@@ -28,10 +29,10 @@ pub fn check_facet_links(ifacet: usize, facets: &[TriangleFacet]) {
...
 
 /// Checks if a convex-hull is properly formed.
-#[cfg(feature = "std")]
-pub fn check_convex_hull(points: &[Point3<Real>], triangles: &[[u32; 3]]) {
-    use std::println;
-
+pub fn check_convex_hull(
+    points: &[Point3<Real>],
+    triangles: &[[u32; 3]],
+) -> Option<ConvexHullError> {
```
Vrixyz, line 35: Returning a `Result` would communicate better the intent.

foxtran: Fixed.

foxtran:
There are couple issues with CI. Will fix them in Sunday or Monday.

PR #357, opened 2025-07-03
Fix zero length 2d segment raycast

### LINKED ISSUE #31: `Segment::intersects_ray` returns false-positive when the segment is zero-length
```rust
use parry2d::math::*;
use parry2d::query::{Ray, RayCast};
use parry2d::shape::Segment;

fn main() {
    // never intersect each other
    let ray = Ray::new(Point::new(0.0, 0.0), Vector::new(1.0, 0.0));
    let segment = Segment {
        a: Point::new(10.0, 10.0),
        b: Point::new(10.0, 10.0),
    };

    // returns true
    let hit = segment.intersects_ray(&Isometry::identity(), &ray, std::f32::MAX);
    assert_eq!(hit, false);
}
```

Vrixyz:
- Fix #31

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Fixed
 
 - Fix `clip_aabb_line` crashing when given incorrect inputs (zero length direction or NAN).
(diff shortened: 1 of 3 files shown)


sebcrozet:
Thank you!
Clippy errors will be addressed in a separate PR.

PR #372, opened 2025-08-13
Fix leaf traversal yielding non-passing leaves

izzymg:
When iterating leaves, the iterator did not check the parent node against the check closure. This caused leaves that did not pass AABB intersection tests to still be yielded during BVH checks, for example.


sebcrozet:
Thank you for this PR! While the proposed fix does work, it will result in most leaf being checked twice (since the iterator already checked them once before they get pushed into the stack).

The actual bug was that the first two nodes (at the root) pushed into the stack were not checked. So if they were leaves, these first two leaves would be left unchecked.

I modified the fix to add the missing check for the root nodes.

sebcrozet:
Clippy will be fixed in #377

Issue #387, opened 2025-10-19
Bug + Fix: Contact between Shape and Composite Shape Issues


dog-molecule:
I'm using Parry via the Rapier physics engine plugin for Godot. I noticed that shapecasts cast against composite shapes were returning bogus results, and on drilling down, I believe I've found the root cause.

It seems that in this case (composite shape/shape contact), an extra inversion is being performed on the vector that defines the relative positions of the two bodies. To illustrate, here's an image from Godot:

<img width="1141" height="417" alt="Image" src="https://github.com/user-attachments/assets/5edc4fa7-f852-42d1-b794-8f2eb66e81b2" />

The circle is casting a shapecast down, which is hitting a composite shape. However, the relative positional vector is being inverted, so the AABB that is being checked for contacts between the two shapes (drawn as a red square here) is exactly opposite (relative to the composite shape's origin) to where it should be. The fix is very simple; it's just a modification to line 58 of src/query/contact/contact_composite_shape_shape.rs:

```{
    CompositeShapeRef(g1)
        //OLD: .contact_with_shape(dispatcher, &pose12.inverse(), g2, prediction) 
        .contact_with_shape(dispatcher, &pose12, g2, prediction) // NEW
        .map(|c| c.1)
}
```

It seems to fix the issue for me, but someone with Parry experience, please take a look at this and let me know if: 
A) you can replicate the issue in a non-Godot-Rapier environment and 
B) if I'm barking up the wrong tree with the solution :)

PR inbound with the fix.

dog-molecule:
PR: https://github.com/dimforge/parry/pull/388

PR #388, opened 2025-10-19
Shape on composite shape contact fix

dog-molecule:
Addresses https://github.com/dimforge/parry/issues/387

(and https://github.com/appsinacup/godot-rapier-physics/issues/394 )


sebcrozet:
Thank you for the fix @dog-molecule! And thank you for the review @Ughuuu.

PR #394, opened 2025-11-07
Fix assertion message

YichiZhang0613:
The assertion's predicate is amount >= 0.0 while message requires amount > 0.0, the message should be changed into "The loosening margin must be non-negative." based on semantic. The same issue elsewhere should also be fixed.
```rust
/// # Arguments
    ///
    /// * `amount` - The amount to increase the radius (must be non-negative)
    ///
    /// # Panics
    ///
    /// Panics if `amount` is negative.
#[inline]
    fn loosened(&self, amount: Real) -> BoundingSphere {
        assert!(amount >= 0.0, "The loosening margin must be positive.");
        BoundingSphere::new(self.center, self.radius + amount)
    }
```


YichiZhang0613:
@sebcrozet  Hi, sorry for the interruption. I see you've been active reviewers recently. Could you please review this PR when you have time? It's been open for two months and I'd appreciate any feedback. Thank you very much.

sebcrozet:
Looks good, thank you!

PR #403, opened 2026-01-19
Fix Bvh::optimize_incremental non-determinism

### LINKED ISSUE #402: Non-determinism in Bvh::optimize_incremental when restoring from serialized snapshot
I originally posted this issue under the Rapier repository, but really the root cause of the issue does come from Parry, and since I want to attempt to fix it I think I should raise an issue here so I can tag against it: https://github.com/dimforge/rapier/issues/910

`Bvh::optimize_incremental` exhibits non-deterministic behavior if you have the type of application that needs to restore from the state of the `Bvh` from a serialized snapshot. For example, in my case, I am creating a multiplayer game engine with rollback netcode and so I often need to restore old snapshots of the game state. 

The reason is that `BvhWorkspace` contains significant state that affects the execution of `Bvh::optimize_incremental`, but `BvhWorkspace` cannot be serialized (which makes sense!). In particular, the `rebuild_frame_index` and `rebuild_start_index` of `BvhWorkspace` are used to make the subtree optimization continue where it left off each time it is called. These fields cannot be snapshotted but are significant to the execution.

My proposed fix is to move these fields into `Bvh` itself. Then they will naturally get serialized/deserialized along with the rest of the `Bvh`. I am going to attempt to make a pull request performing this fix. 

This issue occurs on the latest version of `parry`: `0.26.0`.

raysplaceinspace:
Closes #402 

There is a non-determinism issue in `Bvh::optimize_incremental` when you are restoring your `Bvh` from a serialized snapshot. It occurs because `Bvh::optimize_incremental` relies on significant state in `BvhWorkspace` which is ephemeral and cannot be serialized. This change moves the significant fields into `Bvh` so that they do get serialized, enabling deterministic execution once again.

I've written a lot more background on these two tickets:
- #402 
- https://github.com/dimforge/rapier/issues/910

I have verified that all tests pass, and I have verified that this fixes the non-determinism issue in my application.

I am happy to rework this in other ways if you would prefer.

### DIFF
--- a/src/partitioning/bvh/bvh_optimize.rs
+++ b/src/partitioning/bvh/bvh_optimize.rs
@@ -240,7 +240,8 @@ impl Bvh {
         }
 
         workspace.rebuild_leaves.clear();
-        self.optimization.rebuild_frame_index = self.optimization.rebuild_frame_index.overflowing_add(1).0;
+        self.optimization.rebuild_frame_index =
+            self.optimization.rebuild_frame_index.overflowing_add(1).0;
         let config = self.optimization_config(self.optimization.rebuild_frame_index);
 
         /*
--- a/src/partitioning/bvh/bvh_optimize.rs
+++ b/src/partitioning/bvh/bvh_optimize.rs
@@ -240,15 +240,15 @@ impl Bvh {
         }
 
         workspace.rebuild_leaves.clear();
-        workspace.rebuild_frame_index = workspace.rebuild_frame_index.overflowing_add(1).0;
-        let config = self.optimization_config(workspace.rebuild_frame_index);
+        self.optimization.rebuild_frame_index = self.optimization.rebuild_frame_index.overflowing_add(1).0;
+        let config = self.optimization_config(self.optimization.rebuild_frame_index);
 
         /*
          * Subtree optimizations.
          */
         // let t0 = core::time::Instant::now();
         let num_leaves = self.nodes[0].leaf_count();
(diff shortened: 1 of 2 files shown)


src/partitioning/bvh/bvh_optimize.rs
```diff
@@ -522,6 +522,19 @@ impl Bvh {
     }
 }
 
+/// The optimization state for used by `Bvh::optimize_incremental`.
```
Ughuuu: The optimization state used by `Bvh::optimize_incremental`

raysplaceinspace: Thank you for this! Fixed now!

raysplaceinspace:
@sebcrozet this fixes a determinism issue. I understand that determinism is one of the most important features of Rapier/Parry and so I am wondering if you would be interested in reviewing this change?

Pwnion:
Thanks for your patch @raysplaceinspace, I'm using it for my rollback netcode game as well 🥇

zaxxx:
Thank you @raysplaceinspace, I tried this fix - got zero checksum fails and zero desyncs in my netcode :+1: :tada:

raysplaceinspace:
I'm glad this is helping both of you @Pwnion and @zaxxx! I really hope this change gets merged into Parry and Rapier itself (or an equivalent fix) because it is clearly an issue that affects many of us!

sebcrozet:
Good catch, thank you!

PR #406, opened 2026-03-06
fix: BvhNode::scale now handles negative scale components correctly

robtfm:
check/update bvh for negative scales and remove arg condition. this condition was not mentioned in (at least) `TriMesh::scaled_by`, doesn't seem expensive to check correctly, and is on an infrequent codepath.

--

Previously, scaling an AABB by a negative factor would produce an inverted AABB (mins > maxs on the negated axis), breaking the BVH invariant and causing all intersection/shapecast queries to silently miss geometry.

Fix by taking component-wise min/max of the two scaled corners, so the AABB invariant is preserved regardless of sign. The parent-child containment property of the BVH is maintained since scale(merge(A,B)) == merge(scale(A),scale(B)) holds for any non-zero scale.

This unblocks use of TriMesh::scaled with negative scale (e.g. reflections).

### DIFF
--- a/src/partitioning/bvh/bvh_tree.rs
+++ b/src/partitioning/bvh/bvh_tree.rs
@@ -811,8 +811,12 @@ impl BvhNode {
     /// - [`Bvh::scale`] - Scale an entire BVH tree
     #[inline]
     pub fn scale(&mut self, scale: Vector) {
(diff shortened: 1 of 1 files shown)


sebcrozet:
Thanks!

PR #409, opened 2026-03-24
fix: clean up orphaned BVH nodes when remove creates a partial root

robtfm:
## Summary

- When multiple `Bvh::remove` calls occur without an intervening `refit`, orphaned wide nodes accumulate in the `nodes` vector (by design — the comment says "it will get cleaned up at the next refit"). However, if the root is then reduced to a single leaf (partial root), those orphaned nodes remain, causing `optimize_incremental` to operate on a corrupt tree structure.
- This adds a `truncate(1)` for both `nodes` and `parents` after creating a partial root, since only `node[0]` is reachable in that state.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/partitioning/bvh/bvh_tests.rs
+++ b/src/partitioning/bvh/bvh_tests.rs
@@ -1,6 +1,8 @@
 use crate::bounding_volume::Aabb;
 use crate::math::{Real, Vector};
-use crate::partitioning::{Bvh, BvhBuildStrategy, BvhNode, BvhNodeIndex, TraversalAction};
+use crate::partitioning::{
+    Bvh, BvhBuildStrategy, BvhNode, BvhNodeIndex, BvhWorkspace, TraversalAction,
+};
 
 fn make_test_aabb(i: usize) -> Aabb {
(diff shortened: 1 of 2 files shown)


holmes-software:
Can confirm that this patch fixes my crashes:
```
thread 'Compute Task Pool (1)' (359437) panicked at /home/josh/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/parry3d-0.26.0/src/partitioning/bvh/bvh_traverse_bvtt.rs:44:31:
index out of bounds: the len is 1 but the index is 1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Encountered a panic in system `my_app::plugins::physics::PhysicsSystems::step_rapier`!
Encountered a panic in system `bevy_app::main_schedule::FixedMain::run_fixed_main`!
Encountered a panic in system `bevy_time::fixed::run_fixed_main_schedule`!
Encountered a panic in system `bevy_app::main_schedule::Main::run_main`!
```

I hope this gets merged soon.

sebcrozet:
Thanks!

PR #418, opened 2026-05-18
Bump glamx to 0.2 (glam 0.32)

stuartparmenter:
Bumps glamx to 0.2 (glam 0.32) so parry's math types line up with Bevy's, eliminating the boundary conversions Avian (and other Bevy-side consumers) currently need.


sebcrozet:
Thanks!

PR #419, opened 2026-05-22
bugfix: missing bvh.scale() in voxel collider

akriegman:
I was getting some weird behavior in Avian3d when I had a voxel collider on an entity with a transform.scale other than 1. Adding this line fixed the problem. `voxel_edition.rs` has the similar function `set_voxel_size` which also calls `self.chunk_bvh.scale(...)`, so it seems like this change is in line with the original intent, and it was probably just an oversight that this line is missing.


sebcrozet:
Thanks!

PR #424, opened 2026-06-06
Fix velocity calculation in shape casting

### LINKED ISSUE #423: Strange velocity calculation in shape_cast::cast_shape()
https://github.com/dimforge/parry/blob/b9d137ff12e4ddd67cc723c8641fbefc33705784/src/query/shape_cast/shape_cast.rs#L272-L284

I think the difference of velocities should be with parenthesis like below

```
let vel12 = pos1.rotation.inverse() * (vel2 - vel1);
```

Kyzzsa:
Fixes #423

This pull request makes a correction to the calculation of relative velocity in the `cast_shapes` function. The change ensures that the velocity difference is computed before applying the rotation, which results in correct relative velocity handling during shape casting.

### DIFF
--- a/src/query/shape_cast/shape_cast.rs
+++ b/src/query/shape_cast/shape_cast.rs
@@ -279,6 +279,6 @@ pub fn cast_shapes(
     options: ShapeCastOptions,
 ) -> Result<Option<ShapeCastHit>, Unsupported> {
     let pos12 = pos1.inv_mul(pos2);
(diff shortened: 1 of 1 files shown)


sebcrozet:
Good catch, thank you!

sebcrozet (approved):

raysplaceinspace:
I just noticed this bug and was about to submit a pull request but you fixed it first :) Great work!

PR #425, opened 2026-06-14
oriented polyline

### LINKED ISSUE #254: Implement pseudo normals for polyline similar to how it's done in 3D for trimesh and heightmapfield
For trimesh and heightmapfield there is currently the TrianglePseudoTriangles. There should be similar for polyline, EdgePseudoTriangles or SegmentPseudoTriangles that will fix ghost collisions/bumps.

cscorley:
closes #254 

adds 2D version of TriMeshFlags::ORIENTED + TriMeshFlags::FixInternalEdges

### DIFF
--- a/src/query/point/point_composite_shape.rs
+++ b/src/query/point/point_composite_shape.rs
@@ -134,17 +134,32 @@ impl<S: TypedCompositeShape> CompositeShapeRef<'_, S> {
 impl PointQuery for Polyline {
     #[inline]
     fn project_local_point(&self, point: Vector, solid: bool) -> PointProjection {
-        CompositeShapeRef(self)
-            .project_local_point(point, Real::MAX, solid)
-            .unwrap_or_else(|| unreachable!())
-            .1
+        let (proj, _) = self.project_local_point_and_get_location(point, solid);
+
+        if solid && proj.is_inside {
+            PointProjection::new(true, point)
+        } else {
+            proj
+        }
     }
 
     #[inline]
+    #[allow(unused_mut)] // Because we need mut in 2D but not in 3D.
     fn project_local_point_and_get_feature(&self, point: Vector) -> (PointProjection, FeatureId) {
-        let (seg_id, (proj, feature)) = CompositeShapeRef(self)
+        let (seg_id, (mut proj, feature)) = CompositeShapeRef(self)
             .project_local_point_and_get_feature(point, Real::MAX)
             .unwrap_or_else(|| unreachable!());
+
+        // A point behind the outward pseudo-normal is inside.
+        #[cfg(feature = "dim2")]
+        if let Some(constraints) = self.segment_normal_constraints(seg_id) {
+            let pseudo_normal = match feature {
+                FeatureId::Vertex(i) => constraints.edges[i as usize],
+                _ => constraints.face,
+            };
+            proj.is_inside = (point - proj.point).dot(pseudo_normal) <= 0.0;
+        }
+
         let polyline_feature = self.segment_feature_to_polyline_feature(seg_id, feature);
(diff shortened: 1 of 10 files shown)


src/shape/polyline.rs
```diff
@@ -177,13 +240,107 @@ impl Polyline {
...
+            *normal = normal.normalize_or_zero();
+        }
+
+        self.pseudo_normals = Some(vertex_normals);
+    }
+
+    /// Pseudo-normals are a 2D-only feature; in 3D this is a no-op. 3D stub so
+    /// `set_flags`/`scaled`/`reverse` compile; mirrors `TriMesh`'s dim2 stub.
+    #[cfg(feature = "dim3")]
+    fn compute_pseudo_normals(&mut self) {}
```
sebcrozet: Should be removed.

src/shape/polyline.rs
```diff
@@ -98,6 +123,10 @@ pub struct Polyline {
     bvh: Bvh,
     vertices: Vec<Vector>,
     indices: Vec<[u32; 2]>,
+    /// Per-vertex outward pseudo-normals, present when [`PolylineFlags::ORIENTED`] is set; contact
+    /// normals are then clamped to one side so the polyline acts as a one-sided surface.
+    pseudo_normals: Option<Vec<Vector>>,
+    flags: PolylineFlags,
```
sebcrozet, line 133: Since pseudo-normals (and the `PolylineFlags` as well) are only relevant in 2d, they should be excluded from 3d using `#[cfg(feature = "dim2")]`.

sebcrozet (changes requested):
Thank you for this PR!

Another modification that is needed is the behavior of the implementation of `PointQuery` for `Polyline` so it mirrors the trimesh behavior. If a `Polyline` is oriented, then a point starting inside of it should be considered to be its own projection.
 
See for example the case of an oriented `TriMesh`: https://github.com/dimforge/parry/blob/ef63da4652203d3f32ce058f109fc47859d7e85b/src/query/point/point_composite_shape.rs#L288-L311

src/query/point/point_composite_shape.rs
```diff
@@ -134,17 +134,32 @@ impl<S: TypedCompositeShape> CompositeShapeRef<'_, S> {
 impl PointQuery for Polyline {
     #[inline]
     fn project_local_point(&self, point: Vector, solid: bool) -> PointProjection {
-        CompositeShapeRef(self)
-            .project_local_point(point, Real::MAX, solid)
-            .unwrap_or_else(|| unreachable!())
-            .1
+        let (proj, _) = self.project_local_point_and_get_location(point, solid);
+
+        if solid && proj.is_inside {
```
cscorley: I chose to use solid here arbitrarily because it made sense to me, but happy to drop to the literal is_inside case only

src/query/point/point_composite_shape.rs
```diff
@@ -134,17 +134,32 @@ impl<S: TypedCompositeShape> CompositeShapeRef<'_, S> {
...
+            PointProjection::new(true, point)
+        } else {
+            proj
+        }
     }
 
     #[inline]
+    #[allow(unused_mut)] // Because we need mut in 2D but not in 3D.
```
cscorley, line 141: I think this function can be written a different way to exclude this, but chose this path so it would be mirror more of how the trimesh implementation solves the issue

cscorley:
I believe I have addressed all feedback -- I tried to follow the Trimesh patterns closely here.  I've left inline comments, too.

Thanks!

sebcrozet (approved):

sebcrozet:
Fantastic, thank you!

Issue #180, opened 2024-02-11
`time_of_impact_support_map_support_map` returns `None` when it shouldn't with very specific inputs


bolshoytoster:
I'm trying to get the TOI between a moving square (`Cuboid`) and a stationary line (`Segment`), so I'm using `time_of_impact_support_map_support_map`. This works perfectly fine most of the time, but I realised that I keep getting some `None`s, when there definetely should be a collision.

These `None`s always appear for certain inputs (the square's initial position/velocity). For example, if the square starts at (15.689464, 54.00709) with velocity (0.0081385, -0.999966), it returns `None`, but if you add or subtract .000001 from any of those values, it returns `Some`.

I have created an MRE:
```rust
fn main() {
	let vel = Vector2::new(0.0081385, -0.999966);
	let segment = Segment::new(Point2::new(1., 1.), Point2::new(124., 1.));
	let cuboid = Cuboid::new(Vector2::repeat(2.));

	// These two work properly
	assert!(
		time_of_impact_support_map_support_map(
			&Isometry2::translation(15.689464, 54.00709),
			&vel,
			&segment,
			&cuboid,
			f32::INFINITY,
			true,
		)
		.is_some()
	);

	assert!(
		time_of_impact_support_map_support_map(
			&Isometry2::translation(15.689466, 54.00709),
			&vel,
			&segment,
			&cuboid,
			f32::INFINITY,
			true,
		)
		.is_some()
	);

	// This one fails
	assert!(
		time_of_impact_support_map_support_map(
			&Isometry2::translation(15.689465, 54.00709),
			&vel,
			&segment,
			&cuboid,
			f32::INFINITY,
			true,
		)
		.is_some()
	);
}
```

This obviously also happens with `time_of_impact`, and also with a `Polyline` instead of a single `Segment`.

I narrowed down the point that the wrong one diverges from the correct ones to two lines in `minkowski_ray_cast` in `src/query/gjk/gjk.rs`:
https://github.com/dimforge/parry/blob/e57762fe55d4dc5e7799d385e712e779d5884663/src/query/gjk/gjk.rs#L350-L351

In the correct versions, at the end of the two lines, `simplex.dimension() == 1`, however, in the wrong version, it is 2, causing the function to return `None` early.

I, however, have no idea how simplexes work; so I couldn't go any further. Hopefully someone else can figure this out. For now, I'll probably have to use rapier.

This issue also exists in ncollide.

dubrowgn:
[...]

PR #430, opened 2026-07-11
fix(gjk): shape-cast TOI error scales with target shape extent

### LINKED ISSUE #429: Shape-cast TOI error scales with target shape extent (stagnation exit returns unrefined lower bound)
`cast_shapes` / `minkowski_ray_cast` returns a time of impact that comes back short by an amount that scales with the target shape's extent. Casting a ball (r = 0.5166) straight down at a cuboid, measured max TOI distance error:

| cuboid half-extent | max TOI error |
|---|---|
| 5 m | 0.246 mm |
| 50 m | 3.64 mm |
| 500 m | **139.45 mm** |

The error is one-sided (always short), pose-discontinuous (deterministic per pose, jumping tick to tick), and the witness data (`point1`/`normal1`) stays exact even when the TOI is wrong.

### Reproduction

Failing test on a branch of my fork (test commit only, no fix — based on current master):

```
git clone https://github.com/vikng-dev/parry && cd parry
git checkout 89eb49a
cargo test -p parry3d --test lib shape_cast_toi_accuracy
#   FAILS — prints the error table above
```

### Mechanism

At first I suspected the relative convergence tolerance (gjk.rs:775-786), but it never fires for this case — instrumented over the test's 2,400 casts: 0 hits. What actually happens: `minkowski_ray_cast` advances a lower TOI bound `ltoi` and tracks an upper bound. With large support coordinates orthogonal to the cast direction, float cancellation keeps the upper bound from decreasing; the "upper bounds inconsistencies" last-chance exit fires (gjk.rs:712-715) and returns the current `ltoi` as the hit (gjk.rs:721) — still short by the stagnated gap. All 2,278 erroneous TOIs in the instrumented run exited through this path.

### Impact

Hit this in a 64 Hz networked tank sim: wheel sphere-casts against a 1000 m ground cuboid came back short by p50 33 mm / p99 134 mm / max 200 mm over 19,828 samples; through a 551 kN/m suspension spring that's 10-40 kN of per-tick force noise — a sustained at-rest hull limit cycle (~12 mm heave) and a standing client/server divergence amplifier. Looks like #414 hit the same exit from a Rapier `DynamicShapeCastVehicleController` (wheels vs a 10,000-unit ground backstop, TOI ~10× short) before self-closing. #180's knife-edge `None` returns may live in the same exit structure too.

I have a fix ready — PR follows right after this issue.

vikng-dev:
Fixes #429.

### What

`cast_shapes` / `minkowski_ray_cast` returns a time of impact that comes back short by an amount that scales with the target shape's extent. Casting a ball (r = 0.5166) straight down at a cuboid, measured max TOI distance error:

| cuboid half-extent | before | after (this PR) |
|---|---|---|
| 5 m | 0.246 mm | 0.113 mm |
| 50 m | 3.64 mm | 0.0023 mm |
| 500 m | **139.45 mm** | 0.00024 mm |

The witness data (`point1`/`normal1`) is exact even when the TOI is wrong.

### Reproduction

```
git checkout 89eb49a   # test commit only, no fix
cargo test -p parry3d --test lib shape_cast_toi_accuracy
#   FAILS — prints the error table above (139 mm at 500 m half-extent)
git checkout fix/shape-cast-stagnation-refine
cargo test -p parry3d --test lib shape_cast_toi_accuracy   # passes
```

### Mechanism

At first I suspected the relative convergence tolerance (gjk.rs:775-786), but it never fires for this case — instrumented over the test's 2,400 casts: 0 hits. What actually happens: `minkowski_ray_cast` advances a lower TOI bound `ltoi` and tracks an upper bound. With large support coordinates orthogonal to the cast direction, float cancellation keeps the upper bound from decreasing; the "upper bounds inconsistencies" last-chance exit fires (gjk.rs:712-715) and returns the current `ltoi` as the hit (gjk.rs:721) — still short by the stagnated gap. All 2,278 erroneous TOIs in the instrumented run exited through this path.

### Fix

On the last-chance path, refine `ltoi` from the simplex witnesses already in hand before accepting the impact: project the witness separation along the cast direction (`(witness1 - witness2 - ray.origin) · dir`). The witnesses stay precise exactly where the upper bound stagnates, so this recovers the accuracy the bound lost. Witness and normal generation are untouched — the returned contact data is bit-identical to before; only the TOI scalar changes. Added cost is confined to the rare stagnation exit (≤ dim+1 weighted witness accumulations via the existing `result()`).

Two intentional behavior changes to be aware of:

1. The refined TOI is no longer a certified lower bound: measured overshoot up to +0.113 mm at 5 m half-extent (curvature converts lateral witness deviation into forward error ≈ r·θ²/2), where the old exit was strictly short. Three orders of magnitude below the old error, but callers relying on "TOI never long" should know.
2. A stagnation hit whose refined TOI exceeds `max_time_of_impact` now returns `None`, where the old code returned a wrongly-short hit inside the limit.

### Real-world impact

Found in a 64 Hz networked tank sim: wheel sphere-casts against a 1000 m ground cuboid came back short by p50 33 mm / p99 134 mm / max 200 mm over 19,828 samples; through a 551 kN/m suspension spring that was 10-40 kN of per-tick force noise — a sustained at-rest hull limit cycle (~12 mm heave) and a standing client/server divergence amplifier. #414 hit the same exit independently from a Rapier `DynamicShapeCastVehicleController` (wheels vs a 10,000-unit ground backstop, TOI ~10× short) — this PR is the same diagnosis with a regression test and the refine kept inside the existing exit. Possibly related: #180's pose-knife-edge `None` returns come from this same function's exit structure, though this PR does not address that case.

### Tests

`shape_cast_toi_accuracy_does_not_scale_with_shape_extent` (tests/geometry/time_of_impact3.rs) pins max error ≤ 0.2 mm across half-extents 5/50/500 m and exact witnesses. Full workspace `--tests` suites pass for parry3d, parry2d, parry3d-f64, parry2d-f64; `cargo fmt --check` clean.

### DIFF
--- a/crates/parry3d/tests/geometry/time_of_impact3.rs
+++ b/crates/parry3d/tests/geometry/time_of_impact3.rs
@@ -63,7 +63,9 @@ fn ball_cuboid_toi() {
 #[test]
 fn shape_cast_toi_accuracy_does_not_scale_with_shape_extent() {
     const BALL_RADIUS: Real = 0.5166;
-    const MAX_ERROR: Real = 1.0e-4; // 0.1 mm.
+    const MAX_TOI_ERROR: Real = 2.0e-4; // 0.2 mm.
+    const MAX_WITNESS_ERROR: Real = 1.0e-6;
+    const MAX_NORMAL_ERROR: Real = 5.0e-4;
 
     let ball = Ball::new(BALL_RADIUS);
     let direction = -Vector::Y;
@@ -74,6 +76,7 @@ fn shape_cast_toi_accuracy_does_not_scale_with_shape_extent() {
         let ground_pose = Pose::translation(0.0, -0.5, 0.0);
         let mut max_toi_error: Real = 0.0;
         let mut max_witness_error: Real = 0.0;
+        let mut max_normal_error: Real = 0.0;
 
         // Sweep sub-millimetre start offsets at several lateral positions. This mirrors
         // suspension probes near their resting pose while exercising large support points.
@@ -98,12 +101,17 @@ fn shape_cast_toi_accuracy_does_not_scale_with_shape_extent() {
 
                 let witness1 = ground_pose.transform_point(hit.witness1);
                 max_witness_error = max_witness_error.max(witness1.y.abs());
+                max_normal_error = max_normal_error.max((hit.normal1 - Vector::Y).length());
             }
         }
 
         assert!(
-            max_witness_error <= MAX_ERROR,
-            "shape-cast witness error {max_witness_error} exceeded {MAX_ERROR} for half-extent {half_extent}",
+            max_witness_error <= MAX_WITNESS_ERROR,
+            "shape-cast witness error {max_witness_error} exceeded {MAX_WITNESS_ERROR} for half-extent {half_extent}",
+        );
+        assert!(
+            max_normal_error <= MAX_NORMAL_ERROR,
+            "shape-cast normal error {max_normal_error} exceeded {MAX_NORMAL_ERROR} for half-extent {half_extent}",
         );
         println!(
             "half-extent {half_extent}: max TOI error = {} mm, max witness error = {} mm",
@@ -115,8 +123,8 @@ fn shape_cast_toi_accuracy_does_not_scale_with_shape_extent() {
 
     for (half_extent, max_toi_error) in toi_errors {
         assert!(
-            max_toi_error <= MAX_ERROR,
-            "shape-cast TOI error {max_toi_error} exceeded {MAX_ERROR} for half-extent {half_extent}",
+            max_toi_error <= MAX_TOI_ERROR,
+            "shape-cast TOI error {max_toi_error} exceeded {MAX_TOI_ERROR} for half-extent {half_extent}",
         );
     }
 }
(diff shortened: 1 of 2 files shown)


Ughuuu:
In the test you added, before and after what is the error? And/or at extreme distances, what is the error? (I see in example you put before it was 139mm, and after? 0mm or?)

And this seems to change the gjk algorithm (for the case of error? So it doesn't converge enough or?), is this the only source of error? (eg. there is also a path with epa for eg. in some cases as I remember).

EDIT:

Oh, ok, so after your PR 0.00024 mm error, before 139mm error.

vikng-dev:
Aye, 139mm -> 0.00024mm.
GJK iteration is unchanged, the change only touches the existing "upper bounds inconsistencies" (gjk.rs:712). The problem isn't convergence, in my tests the convergence tolerance never fired - all the wrong TOIs came from the stagnation exit.
epa isn't on this path, cast_shapes is pure gjk. cast that starts in penetration returns TOI = 0 so it's never reached.

sebcrozet:
Thank you! This is superseded by #437
Please reopen if the issue persists.

PR #437, opened 2026-08-07
Triage and fix easy issues

### LINKED ISSUE #17: Bug: very small vertex input to TriMesh results in panic: unwrap on None when distance_to_local_point is called
Issue occurred after converting vertex input to Angstroms, resulting in very small vertex list:

```
vertices = [
  [-0.5, -0.5, 0.5],
  [0.5, -0.5, 0.5],
  [-0.5, 0.5, 0.5],
  [0.5, 0.5, 0.5],
  [-0.5, 0.5, -0.5],
  [0.5, 0.5, -0.5],
  [-0.5, -0.5, -0.5],
  [0.5, -0.5, -0.5],
]
indices = [
  [3, 1, 0],
  [2, 3, 0],
  [5, 3, 2],
  [4, 5, 2],
  [7, 5, 4],
  [6, 7, 4],
  [1, 7, 6],
  [0, 1, 6],
  [5, 7, 1],
  [3, 5, 1],
  [2, 0, 6],
  [4, 2, 6]
]
```

If `length_unit` below is 1E-10 (1 Angstrom):
```
let points = input.vertices.iter().map(|p| Point::new(p[0]*length_unit as f32, p[1]*length_unit as f32, p[2]*length_unit as f32)).collect();
let trimesh = TriMesh::new(points, input.indices.clone());
```
When calling `distance_to_local_point(&p, true)`, parry3d-f32 panics at:
```
thread 'main' panicked at 'called Option::unwrap() on a None value', C:\Users\Jarat\.cargo\registry\src\github.com-1ecc6299db9ec823\parry3d-0.2.0\src\query\point\point_composite_shape.rs:126:59
```

Debug output from failing quadtree below:
[quadtree_debug.txt](https://github.com/dimforge/parry/files/6144769/quadtree_debug.txt)

### LINKED ISSUE #26: `time_of_impact` query no longer has threshold distance
The comment for the `time_of_impact` query suggests the existence of a threshold distance but this has been removed since release 0.3. The old API was useful to me. Is there an alternative way to achieve the same functionality?
```
/// Computes the smallest time when two shapes under translational movement are separated by a
/// distance smaller or equal to `distance`.
///
/// Returns `0.0` if the objects are touching or penetrating.
pub fn time_of_impact(
    pos1: &Isometry<Real>,
    vel1: &Vector<Real>,
    g1: &dyn Shape,
    pos2: &Isometry<Real>,
    vel2: &Vector<Real>,
    g2: &dyn Shape,
    max_toi: Real,
) -> Result<Option<TOI>, Unsupported> {
    let pos12 = pos1.inv_mul(pos2);
    let vel12 = pos1.inverse_transform_vector(&(vel2 - vel1));
    DefaultQueryDispatcher.time_of_impact(&pos12, &vel12, g1, g2, max_toi)
}
```

### LINKED ISSUE #70: False negatives in capsule-cuboid intersection tests
Sweeping a capsule through a cuboid with repeated intersection tests seems to produce multiple false negatives:

```rust
use parry3d::{
    na::{Isometry3, Translation3, Unit, Vector3},
    query::intersection_test,
    shape::{Ball, Capsule, Cuboid, HalfSpace},
};

fn main() {
    let capsule = Capsule::new([0.0, -0.5, 0.0].into(), [0.0, 0.5, 0.0].into(), 0.5);

    // This capsule, equivalent to the ball, also produces false negatives
    // let capsule = Capsule::new([0.0, 0.0, 0.0].into(), [0.0, 0.0, 0.0].into(), 0.5);

    let ball = Ball::new(0.5);

    let halfspace = HalfSpace::new(Unit::new_normalize(Vector3::from([0.0, 1.0, 0.0])));

    // Upper face of the cuboid is coplanar with the outer face of the halfspace
    let cuboid = Cuboid::new([50.0, 50.0, 50.0].into());
    let cuboid_pos = Isometry3 {
        translation: Translation3::from(Vector3::from([0.0, -50.0, 0.0])),
        ..Default::default()
    };

    let steps = 200;
    let y_max = 0.5;
    let y_min = -0.5;
    let step_size = (y_max - y_min) / steps as f32;

    let mut capsule_cuboid = 0;
    let mut capsule_halfspace = 0;
    let mut ball_cuboid = 0;
    let mut ball_halfspace = 0;

    for step in 0..steps {
        let y = y_min + step_size * step as f32;

        let test_pos = Isometry3 {
            translation: Translation3::from([0.0, y, 0.0]),
            ..Default::default()
        };

        if intersection_test(&test_pos, &capsule, &Isometry3::default(), &halfspace).unwrap() {
            capsule_halfspace += 1;
        }

        if intersection_test(&test_pos, &capsule, &cuboid_pos, &cuboid).unwrap() {
            capsule_cuboid += 1;
        }

        if intersection_test(&test_pos, &ball, &Isometry3::default(), &halfspace).unwrap() {
            ball_halfspace += 1;
        }

        if intersection_test(&test_pos, &ball, &cuboid_pos, &cuboid).unwrap() {
            ball_cuboid += 1;
        }
    }

    println!("capsule-cuboid intersections:    {}/{}", capsule_cuboid, steps);
    println!("capsule-halfspace intersections: {}/{}", capsule_halfspace, steps);
    println!("ball-cuboid intersections:       {}/{}", ball_cuboid, steps);
    println!("ball-halfspace intersections:    {}/{}", ball_halfspace, steps);
}

```

Outputs:

```
capsule-cuboid intersections:    86/200
capsule-halfspace intersections: 200/200
ball-cuboid intersections:       200/200
ball-halfspace intersections:    200/200
```

### LINKED ISSUE #76: Distance between a point and a degenerate triangle computed incorrectly
I have a triangle and a point for which I need to compute the distance. However, when the triangle is degenerate, the result can be very very wrong. Here's a test to demonstrate the issue:

```
        let p = na::Point3::new(1.10000002, -7.9000001, 16.5879993);

        let a = na::Point3::new(2.27699995, -7.9000001, 16.3180008);
        let b = na::Point3::new(-0.569999993, -8.10000038, 16.6070004);
        let c = na::Point3::new(-0.569999993, -8.10000038, 16.6070004);

        let line = rapier3d::parry::shape::Polyline::new(vec![a, b], None);

        let tri = rapier3d::parry::shape::Triangle::new(a, b, c);

        assert_eq!(tri.area(), 0.0);

        assert!(tri.local_aabb().contains_local_point(&p));

        let tri_dist = rapier3d::parry::query::PointQuery::distance_to_local_point(&tri, &p, true);
        let line_dist = rapier3d::parry::query::PointQuery::distance_to_local_point(&line, &p, true);
        
        assert!(line_dist > 0.0);
        assert_eq!(tri_dist, line_dist);

        // :tri_pt_dist_degen' panicked at 'assertion failed: `(left == right)`
        // left: `0.0`,
        // right: `0.17147861`'
```

If I treat the triangle as a line, the distance is computed correctly. At least it looks correct visually judging.

I'm using rapier3d 0.11.1, which still installs the old parry3d 0.7.1.0

### LINKED ISSUE #79: closest_points panics when a Cylinder & Trimesh are not within max_dist
[`closest_points`](https://docs.rs/parry3d/latest/parry3d/query/fn.closest_points.html) panics when a `Cylinder` and `Trimesh` are tested and the objects are not within `max_dist` of each other. I believe the function is actually supposed to return `Disjoint` instead. Here is a specific example I've found:

```rust
use parry3d::{math::*, query, shape::*};

fn main() {
    let cylinder = Cylinder::new(0.5, 1.0);

    let vertices = vec![
        Point::new(1.0, -1.0, 0.0),
        Point::new(-1.0, -1.0, 0.0),
        Point::new(0.0, -1.0, 1.0),
    ];
    let indices = vec![[0, 1, 2]];
    let trimesh = TriMesh::new(vertices, indices);

    // let max_dist = 0.5; // Fine
    let max_dist = 0.49; // Panics

    let _ = query::closest_points(
        &Isometry::identity(),
        cylinder.clone_box().as_ref(),
        &Isometry::identity(),
        trimesh.clone_box().as_ref(),
        max_dist,
    );
}
```

### LINKED ISSUE #106: Inconsistent TOI between ball-vs-ball and support-map-vs-support-map
In v0.11.1, time-of-impact queries are returning inconsistent results for `toi=0`, depending on which shapes are involved.

If both shapes are balls, `toi=0` results in `status=converged` (https://github.com/dimforge/parry/blob/v0.11.1/src/query/time_of_impact/time_of_impact_ball_ball.rs#L45)

```rust
	 let status = if inside && center.coords.norm_squared() < rsum * rsum {
	 	 TOIStatus::Penetrating
	 } else {
	 	 TOIStatus::Converged
	 };
```

If one shape is a cuboid, and the other a ball, `toi=0` results in `status=pentration` (https://github.com/dimforge/parry/blob/v0.11.1/src/query/time_of_impact/time_of_impact_support_map_support_map.rs#L50):

```rust
	 status: if toi.is_zero() {
		TOIStatus::Penetrating
	} else {
		TOIStatus::Converged
	},
```

### LINKED ISSUE #109: SegmentsIntersection::Segment::first_loc1 is OnEdge([1.0, 0.0]) instead of OnVertex(0)
When using `segments_intersection2d`, I get misleading results. Instead of `first_loc1` being `OnVertex(0)`, I get `OnEdge([1.0, 0.0])`. While they represent the same point, the result is misleading since the doc comment of OnEdge says 
`The point lies on the segment interior.`, and `OnVertex` would be more precise.

When I run the following code:

```
use rapier2d::prelude::*;

let seg1 = Segment::new(point![10.0, 0.0], point![10.0, 10.0]);
let seg2 = Segment::new(point![10.0, 0.0], point![10.0, 10.0]);

let intersection = rapier2d::parry::utils::segments_intersection2d(&seg1.a, &seg1.b, &seg2.a, &seg2.b, 0.0).unwrap();

let rapier2d::parry::utils::SegmentsIntersection::Segment { first_loc1, first_loc2, second_loc1, second_loc2 } = intersection else {
    unreachable!("The intersection should be a Segment intersection!");
};

dbg!(first_loc1);
dbg!(first_loc2);
dbg!(second_loc1);
dbg!(second_loc2);
  ```

The output is

```
first_loc1 = OnEdge([1.0, 0.0])
first_loc2 = OnVertex(0)
second_loc1 = OnEdge([0.0, 1.0,])
second_loc2 = OnVertex(1)
```

The output should be:
```
first_loc1 =OnVertex(0)
first_loc2 = OnVertex(0)
second_loc1 = OnVertex(1)
second_loc2 = OnVertex(1)
```

rustc: 1.65
parry2d: tested on 0.10 and 0.11.1

### LINKED ISSUE #111: Triangle::area wildly inaccurate for degenerate triangles (compute in f64 please)
When I compute the area of a triangle with the provided `Triangle::area` -function, the area can be very wrong.

### LINKED ISSUE #157: Frustum-frustum intersection jitter

### LINKED ISSUE #168: [Bug?] Contact of circle and polyline at its vertex produces 0 normal
Not sure if this is something fixable as I'm not familiar with the math, but (seemingly) when a circle touches a line just at its vertex, a normal of 0 is produced.

In the case of a one-way platform, this can cause spurious collisions on the very ends of a platform when pass-through is supposed to happen.

### LINKED ISSUE #177: Implement Debug for TriMesh
This would be useful to quickly validate if a TriMesh looks as expected (and to allow derives on types containing trimeshes).

The naive derived Debug impl is obviously unhelpful. Instead, perhaps report:

- The number of triangles
- The AABB
- Any flags
- The topology information, if present

### LINKED ISSUE #180: `time_of_impact_support_map_support_map` returns `None` when it shouldn't with very specific inputs

### LINKED ISSUE #193: Shape casting returns incorrect normals
Coming from [bevy_xpbd](https://github.com/Jondolf/bevy_xpbd), normals are very imprecise when using shape casts via bevy_xpbd's spacial query's shape_hits function. i am creating a kinematic character controller, and it gives incorrect values when moving almost orthogonal to colliders, even when they are completely aligned to grid. it isn't by a ton, but it's enough to clip through colliders. after discussing it with the creator of bevy_xpbd and a mod of the bevy discord, i was told it was likely a parry issue and to report it here.

i've also noticed that normals are almost never returned normalized. there tends to be a bit of fluff in one axis or the other, even when colliding head on.

if you need anything else from me, please let me know.

in the image i was clipping into a 10x5x0.1 (bevy coordinates) meter wall aligned to grid. collisions work fine when pushing directly against the wall, but fail when moving into it at a shallow angle.

image

### LINKED ISSUE #215: function hang infinitely

### LINKED ISSUE #252: Consider using `polyhedral_mass_properties`
[`polyhedral_mass_properties`](https://docs.rs/polyhedral_mass_properties) is a library that I just published to calculate mass properties (mass, center of mass and inertia matrix/tensor) of triangle meshes. It is at least 2x faster than [`MassProperties::from_trimesh`](https://docs.rs/parry3d-f64/latest/parry3d_f64/mass_properties/struct.MassProperties.html#method.from_trimesh). See [benchmark](https://codeberg.org/mo8it/polyhedral_mass_properties/src/commit/94124419395d3986c9a6ab728ab10bb999a0a250/benches/algorithm.rs#L90-L112).

Using [`MassProperties::reconstruct_inertia_matrix`](https://docs.rs/parry3d-f64/latest/parry3d_f64/mass_properties/struct.MassProperties.html#method.reconstruct_inertia_matrix) returns a matrix where `xy` and `xz` are swapped. This is probably a bug in the implementation of Parry (I compared with other implementations, not only mine).

If you need any help, please let me know. I could add a version for `f32` and add the Apache license.

### LINKED ISSUE #260: Bounding sphere merge sometimes results in non including both shapes
Due to float imprecisions, we can't be certain a call to `bounding1.merged(bounding2)` will results in a bounding containing both shapes.

<details><summary>Reproduction</summary>
<p>

```rs
        let cube1_pos = Vec2::new(3.75, 2.6516504);
        let cube1_pos = Isometry2::translation(cube1_pos.x, cube1_pos.y);
        let cube2_pos = Isometry2::identity();
        dbg!(cube1_pos.translation);
        dbg!(cube2_pos.translation);

        let bounding_sphere_cube1 = cube1.bounding_sphere(&cube1_pos);
        let bounding_sphere_cube2 = cube2.bounding_sphere(&cube2_pos);

        // Merge the two spheres.
        let bounding_bounding_sphere = dbg!(bounding_sphere_cube1.merged(&bounding_sphere_cube2));

        // This fails /!\
        assert!(bounding_bounding_sphere.contains(&bounding_sphere_cube2));
```

```
[crates/parry2d/examples/bounding_sphere2d.rs:31:9] cube1_pos.translation = [
    3.75,
    2.6516504,
]
[crates/parry2d/examples/bounding_sphere2d.rs:32:9] cube2_pos.translation = [
    0.0,
    0.0,
]
[crates/parry2d/examples/bounding_sphere2d.rs:38:40] bounding_sphere_cube1.merged(&bounding_sphere_cube2) = BoundingSphere {
    center: [
        1.7072396,
        1.2072008,
    ],
    radius: 3.208967,
}
thread 'main' panicked at crates/parry2d/examples/bounding_sphere2d.rs:39:9:
assertion failed: bounding_bounding_sphere.contains(&bounding_sphere_cube2)
```

</p>
</details> 

That's why `bounding.loosened()` exists.

The examples shouldn´t encourage this assumption (https://github.com/dimforge/parry/blob/837291f1a051dc04e6e4ab8bc2e5b438ce66d257/crates/parry2d/examples/bounding_sphere2d.rs#L31)

### LINKED ISSUE #274: `parry::query::contact` Does not work with heightfield
Using capsule vs heightfield results in `Err::Unsupported`. This causes the godot-rapier character controller to just fall through the heightfield: https://github.com/appsinacup/godot-rapier-physics/issues/236

### LINKED ISSUE #315: parry3d::query::contact
For the following 2 boxes, they touch on plane x=1

```rs
use nalgebra::Vector3 as v3;

let box1 = Cuboid::new(v3::new(0.5, 0.5, 0.5));
let pos1 = Isometry3::translation(0.5, 0.5, 0.5);
let pos2 = Isometry3::translation(1.5, 0.5, 0.5);
let exp_contact = contact(&pos1, &box1, &pos2, &box1, 0.0).unwrap().unwrap();

assert!(true == false, "pt1: {:?}, pt2:{:?}", exp_contact.point1, exp_contact.point2);
```
They are pt1: [0.5, 0.5, 0.5], pt2:[0.5, 0.5, 0.5], definitely wrong

And normal1 is [0.0, -1.0, 0.0], which I doubt is also wrong.

### LINKED ISSUE #332: Sensors and large heightfields are slow together
In my game I have a decently sized heightmap representing the ground and a player around which I want to put a sensor, in order to detect when it is near things. Adding that sensor makes things terribly slow though, to the point where it feels like it could be a bug.

Repro (`rapier3d = "0.12.0"`)
```rust
use rapier3d::{
    na::{Dynamic, VecStorage},
    prelude::*,
};

fn main() {
    let mut rigid_body_set = RigidBodySet::new();
    let mut collider_set = ColliderSet::new();
    const HEIGHTFIELD_SIDE: f32 = 256.0;

    {
        const HEIGHTFIELD_NUM_SUBDIVS: usize = 2048;
        let heights: Vec<f32> = vec![1.0; HEIGHTFIELD_NUM_SUBDIVS * HEIGHTFIELD_NUM_SUBDIVS];
        let rigid_body_handle = rigid_body_set.insert(RigidBodyBuilder::fixed());
        let collider = ColliderBuilder::heightfield(
            DMatrix::from_data(VecStorage::new(
                Dynamic::new(HEIGHTFIELD_NUM_SUBDIVS),
                Dynamic::new(HEIGHTFIELD_NUM_SUBDIVS),
                heights,
            )),
            vector!(HEIGHTFIELD_SIDE, 1.0, HEIGHTFIELD_SIDE),
        )
        .translation(0.5 * vector!(HEIGHTFIELD_SIDE, 0.0, HEIGHTFIELD_SIDE))
        .build();
        collider_set.insert_with_parent(collider, rigid_body_handle, &mut rigid_body_set);
    }

    let player_body_handle = {
        let rigid_body_handle = rigid_body_set.insert(
            RigidBodyBuilder::dynamic()
                .translation(vector!(0.5 * HEIGHTFIELD_SIDE, 4.0, 0.5 * HEIGHTFIELD_SIDE))
                .lock_rotations()
                .build(),
        );

        let player_collider = ColliderBuilder::capsule_y(0.5, 0.5)
            .restitution(0.0)
            .build();
        collider_set.insert_with_parent(player_collider, rigid_body_handle, &mut rigid_body_set);

        // This sensor makes the simulation slow.
        let player_sensor = ColliderBuilder::ball(1.1).sensor(true).build();
        collider_set.insert_with_parent(player_sensor, rigid_body_handle, &mut rigid_body_set);

        rigid_body_handle
    };

    let gravity = vector![0.0, -9.81, 0.0];
    let integration_parameters = IntegrationParameters::default();
    let mut physics_pipeline = PhysicsPipeline::new();
    let mut island_manager = IslandManager::new();
    let mut broad_phase = BroadPhase::new();
    let mut narrow_phase = NarrowPhase::new();
    let mut impulse_joint_set = ImpulseJointSet::new();
    let mut multibody_joint_set = MultibodyJointSet::new();
    let mut ccd_solver = CCDSolver::new();

    for _ in 0..50 {
        let iter_start = std::time::Instant::now();
        physics_pipeline.step(
            &gravity,
            &integration_parameters,
            &mut island_manager,
            &mut broad_phase,
            &mut narrow_phase,
            &mut rigid_body_set,
            &mut collider_set,
            &mut impulse_joint_set,
            &mut multibody_joint_set,
            &mut ccd_solver,
            &(),
            &(),
        );
        let player_body = &rigid_body_set[player_body_handle];
        println!(
            "Iter time: {:?}, player.y: {}",
            iter_start.elapsed(),
            player_body.translation().y
        );
    }
}
```

Output
```
...
Iter time: 141.909µs, player.y: 2.378625
Iter time: 170.593µs, player.y: 2.2832499
Iter time: 127.047µs, player.y: 2.18515
Iter time: 21.046758119s, player.y: 2.0843248  <---- this is where the player sensor hits the ground
Iter time: 21.025214909s, player.y: 1.9807749
Iter time: 20.945551641s, player.y: 1.99394
...
```

### LINKED ISSUE #345: 3D Voxels example crash
I encountered this crash on voxels example, it reproduces on mac + PopOS after a few seconds:

The crash happens at https://github.com/dimforge/parry/blob/025c77d2ee438dde2c9dfc8cd81f60cceaab35e4/src/shape/voxels.rs#L858.

`rel_key.z` is somehow -1, so when cast to u32, that leads to an overflowing multiplication.

The `point_key.z` in argument is under the `min_dimension.z`:
```rs
voxel_key = [
    126,
    0,
    -1,
]
```
```rs
self.domain_mins = [
    0,
    -11,
    0,
]
```

So either the point_key is wrongly computed, or the dimension.

<details><summary>Crash</summary>
<p>

```
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.16s
     Running `target/debug/all_examples3`
2025-06-12T07:57:42.318205Z  INFO bevy_render::renderer: AdapterInfo { name: "AMD Radeon RX 7600 (RADV NAVI33)", vendor: 4098, device: 29824, device_type: DiscreteGpu, driver: "radv", driver_info: "Mesa 24.0.3-1pop1~1711635559~22.04~7a9f319", backend: Vulkan }
2025-06-12T07:57:42.375048Z  INFO bevy_winit::system: Creating new window "Rapier: 3D demos" (0v1#4294967296)
2025-06-12T07:57:42.375078Z  INFO winit::platform_impl::linux::x11::window: Guessed window scale factor: 1

[/home/tb/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/parry3d-0.21.1/src/shape/voxels.rs:858:9] rel_key = [
    [
        126,
        11,
        -1,
    ],
]

thread 'main' panicked at /home/tb/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/parry3d-0.21.1/src/shape/voxels.rs:858:9:
attempt to add with overflow
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_const::panic_const_add_overflow
             at /rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/core/src/panicking.rs:178:21
   3: parry3d::shape::voxels::Voxels::linear_index
             at /home/tb/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/parry3d-0.21.1/src/shape/voxels.rs:858:9
   4: parry3d::query::contact_manifolds::contact_manifolds_voxels_shape::CanonicalVoxelShape::from_voxel
             at /home/tb/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/parry3d-0.21.1/src/query/contact_manifolds/contact_manifolds_voxels_shape.rs:355:17
   5: parry3d::query::contact_manifolds::contact_manifolds_voxels_shape::contact_manifolds_voxels_shape
             at /home/tb/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/parry3d-0.21.1/src/query/contact_manifolds/contact_manifolds_voxels_shape.rs:145:26
   6: parry3d::query::contact_manifolds::contact_manifolds_voxels_shape::contact_manifolds_voxels_shape_shapes
             at /home/tb/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/parry3d-0.21.1/src/query/contact_manifolds/contact_manifolds_voxels_shape.rs:87:9
   7: <parry3d::query::default_query_dispatcher::DefaultQueryDispatcher as parry3d::query::query_dispatcher::PersistentQueryDispatcher<ManifoldData,ContactData>>::contact_manifolds
             at /home/tb/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/parry3d-0.21.1/src/query/default_query_dispatcher.rs:555:21
   8: rapier3d::geometry::narrow_phase::NarrowPhase::compute_contacts::{{closure}}
             at ./crates/rapier3d/../../src/geometry/narrow_phase.rs:960:25
   9: <core::slice::iter::IterMut<T> as core::iter::traits::iterator::Iterator>::for_each
             at /home/tb/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/iter/macros.rs:277:21
  10: rapier3d::geometry::narrow_phase::NarrowPhase::compute_contacts
             at ./crates/rapier3d/../../src/geometry/narrow_phase.rs:820:9
  11: rapier3d::pipeline::physics_pipeline::PhysicsPipeline::detect_collisions
             at ./crates/rapier3d/../../src/pipeline/physics_pipeline.rs:150:9
  12: rapier3d::pipeline::physics_pipeline::PhysicsPipeline::step
             at ./crates/rapier3d/../../src/pipeline/physics_pipeline.rs:627:13
  13: rapier_testbed3d::harness::Harness::step_with_graphics
             at ./crates/rapier_testbed3d/../../src_testbed/harness/mod.rs:235:9
  14: rapier_testbed3d::testbed::update_testbed
             at ./crates/rapier_testbed3d/../../src_testbed/testbed.rs:1436:17
  15: core::ops::function::FnMut::call_mut
             at /home/tb/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:166:5
  16: core::ops::function::impls::<impl core::ops::function::FnMut<A> for &mut F>::call_mut
             at /home/tb/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:294:13
  17: <Func as bevy_ecs::system::function_system::SystemParamFunction<fn(F0,F1,F2,F3,F4,F5,F6,F7,F8,F9,F10,F11,F12,F13) .> Out>>::run::call_inner
             at /home/tb/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/bevy_ecs-0.15.2/src/system/function_system.rs:1002:21
  18: <Func as bevy_ecs::system::function_system::SystemParamFunction<fn(F0,F1,F2,F3,F4,F5,F6,F7,F8,F9,F10,F11,F12,F13) .> Out>>::run
             at /home/tb/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/bevy_ecs-0.15.2/src/system/function_system.rs:1005:17
  19: <bevy_ecs::system::function_system::FunctionSystem<Marker,F> as bevy_ecs::system::system::System>::run_unsafe
             at /home/tb/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/bevy_ecs-0.15.2/src/system/function_system.rs:800:19
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Encountered a panic in system `rapier_testbed3d::testbed::update_testbed`!
Encountered a panic in system `bevy_app::main_schedule::Main::run_main`!
```

</p>
</details>

### LINKED ISSUE #373: Wrong witness1 value for ShapeCastHit in cast_shapes_voxels_shape
https://github.com/dimforge/parry/blob/d955f7527847997ea6a1f4c42602ff9b4a8669d3/src/query/shape_cast/shape_cast_voxels_shape.rs#L23-L42

The returned witness1 should be in local space for the voxel shape. However, we delegate the shape cast to a cuboid representing the individual voxel and never transform the result from this hit back into voxel-shape space.

### LINKED ISSUE #382: `Voxels#voxels()` and `voxels_in_range()` no longer return empty voxels
I believe the BVH chunking already skips empty sections and when we drop into our `VoxelsChunkRef#voxels_in_range()` we're dropping empty voxels.

I'm not advocating for reverting this behavior, only returning the non-empty voxels is probably a better API, we'll just want to update the documentation.

### LINKED ISSUE #387: Bug + Fix: Contact between Shape and Composite Shape Issues

### LINKED ISSUE #395: CompositeShapeRef::project_local_point panics when no match is found
[`CompositeShapeRef::project_local_point`](https://github.com/dimforge/parry/blob/v0.25.1/src/query/point/point_composite_shape.rs#L79) panics when no match is found.

Original issue https://github.com/avianphysics/avian/issues/872.

Essentially `best_val` is never getting set to anything inside `Bvh::find_best` based on my step-through debugging.

https://github.com/dimforge/parry/blob/v0.25.1/src/partitioning/bvh/bvh_traverse.rs#L343

### LINKED ISSUE #396: intersection_test returns Ok(false) for two overlapping cylinders
Hello,
I'm using parry `0.25.3` and and the following snippet returns `Ok(false)` for two overlapping cylinders:
```rust
use parry3d::na::Isometry3;
use parry3d::query::intersection_test;
use parry3d::shape::Cylinder;

fn main() {
    let iso = Isometry3::identity();
    let big = Cylinder::new(50.0, 100.0);
    let small = Cylinder::new(1.5, 1.0);

    let res = intersection_test(&iso, &big, &iso, &small);
    log::info!("direct cylinder test: {res:?}"); // Ok(false)
}
```

Replacing `Cylinder/Cylinder` by `Ball/Ball` or `Capsule/Cylinder` works so I believe this is a bug in the intersection test between two cylinders but I don't really know how to dig deeper.

Edit:
This does not work either:
```rust
fn direct_test() {
    use parry3d::na::Isometry3;
    use parry3d::query::intersection_test;
    use parry3d::shape::Cylinder;

    let iso = Isometry3::identity();
    let big = Cylinder::new(50.0, 100.0);
    let small = Capsule::new(
        Point3::new(-0.0, -50.0, -0.0),
        Point3::new(-0.0, 50.0, -0.0),
        10.0,
    );

    let res = intersection_test(&iso, &big, &iso, &small);
    log::info!("direct cylinder test: {res:?}"); // Ok(false)
}
```

### LINKED ISSUE #399: Mutable BVH
I need to rebuild BVH from scratch every tick is there any chance of adding api to rebuild from leaves? instead of reallocating bvh

### LINKED ISSUE #408: Hertel-Mehlhorn Algorithm constantly calls remove() on a vector
Given that remove() is O(n) in time complexity, could this line of code be replaced by a swap_remove(), or can we simply replace that element by a None value?
https://github.com/dimforge/parry/blob/72f842d9dab8c3729d45ee9b26efe86fc744d1c1/src/transformation/hertel_mehlhorn.rs#L114

### LINKED ISSUE #428: Parry has wrong version selector for spade, leading to compilation errors
Parry uses `ConstrainedDelaunayTriangulation::try_bulk_load_cdt`, which was added in `2.15`. Yet in `Cargo.toml` `spade`-dependency is defined like this

```
spade = { version = "2.9", default-features = false }
```

This can lead to compilation errors if there's already earlier than 2.15 version of spade in the workspace. This happened in my project.

### LINKED ISSUE #429: Shape-cast TOI error scales with target shape extent (stagnation exit returns unrefined lower bound)

### LINKED ISSUE #431: `query::distance` between two cuboids is asymmetric and overestimates the true separation
```rust
use parry2d::math::{Pose, Vector};
use parry2d::query;
use parry2d::query::closest_points::ClosestPoints;
use parry2d::shape::Cuboid;

fn main() {
    let c1 = Cuboid::new(Vector::new(1.0, 1.0));
    let c2 = Cuboid::new(Vector::new(1.0, 2.0));
    let p1 = Pose::identity();
    let p2 = Pose::new(Vector::new(-5.573167, 0.0), 0.0);

    let d12 = query::distance(&p1, &c1, &p2, &c2).unwrap();
    let d21 = query::distance(&p2, &c2, &p1, &c1).unwrap();

    println!("distance(c1, c2) = {}", d12);
    println!("distance(c2, c1) = {}", d21);

    if let ClosestPoints::WithinMargin(a, b) =
        query::closest_points(&p1, &c1, &p2, &c2, f32::MAX).unwrap()
    {
        println!("closest_points distance = {}", (a - b).length());
    }
}
```

Output:

```
distance(c1, c2) = 3.710461
distance(c2, c1) = 3.5731668
closest_points distance = 3.5731668
```

`query::distance` gives a different answer depending on argument order, and the `c1, c2` order overestimates the true gap between the two cuboids (5.573167 − 1 − 1 = 3.5731668, which is what `closest_points` and the `c2, c1` order both agree on).

Tested on `parry2d` 0.29.0.

BTW, this bug was found using [hegel](https://crates.io/crates/hegeltest). Happy to contribute the tests if you're interested.

sebcrozet:
This addresses multiple issues and add non-regression tests for issues that were already fixed in the latest parry version. See each commit individually (one commit per related issue).

All the reproducers/non-regression tests were generated from the issue with AI assistance.
Issues marked as `Fix` include a fix, and those marked as `Close` just have a non-regression test as it appeared already fixed on `main`.

Fix dimforge/rapier#961
Fix dimforge/rapier#969
Fix dimforge/rapier#332
Close dimforge/rapier#168
Fix #429
Fix #396
Close #70
Fix #180
Fix #106
Fix #193
Close #26
Close #157
Fix #111
Fix #76
Fix #109
Fix #395
Close #17
Close #79
Close #215
Fix #177
Close #252
Fix #373
Fix #382
Close #345
Close #387
Close #399
Fix #428
Fix #260
Fix #408
Close #431
Fix #315
Fix #274
Fix dimforge/rapier#810

### DIFF
--- a/.github/workflows/parry-ci-build.yml
+++ b/.github/workflows/parry-ci-build.yml
@@ -57,6 +57,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: sudo apt-get install -y cmake libxcb-composite0-dev
+      # WORKAROUND: zune-core 0.5.2 made its `warn!` macro expand to nothing, which breaks
+      # zune-jpeg (all 0.5.x), where it is called in expression position. Reached through the
+      # kiss3d -> gltf/image dev-dependency. Remove once upstream is fixed.
+      - name: Pin zune-core (0.5.2 breaks zune-jpeg)
+        run: cargo update -p zune-core@0.5 --precise 0.5.1
       - name: Run tests
         run: cargo test --features wavefront
       - name: Run tests (parallel)
@@ -107,5 +112,8 @@ jobs:
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: nightly
+      # See the `tests` job: zune-core 0.5.2 breaks zune-jpeg. Remove once upstream is fixed.
+      - name: Pin zune-core (0.5.2 breaks zune-jpeg)
+        run: cargo update -p zune-core@0.5 --precise 0.5.1
       - name: check benchmarks
         run: cargo +nightly check --benches
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,60 @@
+## Unreleased
+
+### Added
+
+- `Debug` for `TriMesh` now prints a summary instead of just the type name
+  ([#177](https://github.com/dimforge/parry/issues/177)).
+- `query::contact` now supports heightfields, instead of returning `Err(Unsupported)`
+  ([#274](https://github.com/dimforge/parry/issues/274)).
+
+### Fixed
+
+- Fix GJK reporting overlapping axisymmetric shapes as disjoint when the simplex stalls on a
+  degenerate support direction ([#396](https://github.com/dimforge/parry/issues/396)).
+- Fix wrong shape-cast normals and distances for stalled GJK simplices, now classified as touching
+  or penetrating from the solver's own distance bound ([#193](https://github.com/dimforge/parry/issues/193)).
+- Fix shape-casts missing grazing hits at large coordinates, by scaling the GJK ray-cast tolerance
+  with the support magnitude ([#180](https://github.com/dimforge/parry/issues/180)).
+- Fix shape-cast time of impact being short by an amount scaling with the shape's extent
+  ([#429](https://github.com/dimforge/parry/issues/429)).
+- Fix `ShapeCastStatus::PenetratingOrWithinTargetDist` being reported for merely touching support-map
+  casts ([#106](https://github.com/dimforge/parry/issues/106)).
+- Fix `contact` between exactly-touching shapes returning witness points at the shapes' centers
+  ([#315](https://github.com/dimforge/parry/issues/315)).
+- Fix `Triangle::area` returning wildly wrong values for degenerate triangles
+  ([#111](https://github.com/dimforge/parry/issues/111)).
+- Fix point projection on degenerate triangles reporting the point as inside
+  ([#76](https://github.com/dimforge/parry/issues/76)).
+- Fix `segments_intersection2d` classifying exact segment endpoints as `OnEdge` instead of `OnVertex`
+  ([#109](https://github.com/dimforge/parry/issues/109)).
+- Fix composite-shape point projection panicking on non-finite points
+  ([#395](https://github.com/dimforge/parry/issues/395)).
+- Fix `HeightField::project_local_point` iterating on every element, making it O(rows * cols)
+  ([rapier#332](https://github.com/dimforge/rapier/issues/332)).
+- Fix single-point contact manifolds on cylinder and cone caps, by orienting the cap's polygonal
+  approximation toward the contact ([rapier#810](https://github.com/dimforge/rapier/issues/810)).
+- Fix missing speculative contacts between a `Voxels` shape and a shape within the prediction
+  distance ([#404](https://github.com/dimforge/parry/issues/404)).
+- Fix `Voxels` shape-cast `witness1` not being expressed in the voxels shape's local frame
+  ([#373](https://github.com/dimforge/parry/issues/373)).
+- Fix an out-of-bounds panic in the binned BVH builder on degenerate leaf AABBs
+  ([rapier#961](https://github.com/dimforge/rapier/issues/961)).
+- Fix out-of-bounds indices in `to_outline` for round shapes with a zero border radius
+  ([rapier#969](https://github.com/dimforge/rapier/issues/969)).
+- Fix the quadratic polygon removal in `hertel_mehlhorn`
+  ([#408](https://github.com/dimforge/parry/issues/408)).
+- Require `spade` 2.15, the minimum version providing `try_bulk_load_cdt`
+  ([#428](https://github.com/dimforge/parry/issues/428)).
+
+### Modified
+
+- Document the frame conventions of the `ShapeCastHit` fields
+  ([rapier#933](https://github.com/dimforge/rapier/issues/933)).
+- Document that the `Voxels` iterators only yield non-empty voxels
+  ([#382](https://github.com/dimforge/parry/issues/382)).
+- Document that `BoundingVolume::merge`/`merged` don't guarantee strict containment of their inputs
+  ([#260](https://github.com/dimforge/parry/issues/260)).
+
 ## 0.30.1
 
 ### Fixed
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ indexmap = { version = "2", features = ["serde"] }
 hashbrown = { version = "0.17", default-features = false, features = [
     "default-hasher",
 ] }
-spade = { version = "2.9", default-features = false }
+spade = { version = "2.15", default-features = false }
 rayon = "1"
 bytemuck = { version = "1", features = ["derive"] }
 log = "0.4"
--- a/crates/parry2d/tests/issue_109_segments_intersection_endpoints.rs
+++ b/crates/parry2d/tests/issue_109_segments_intersection_endpoints.rs
@@ -1,0 +1,94 @@
+// Regression test for https://github.com/dimforge/parry/issues/109
+//
+// `segments_intersection2d` used to classify exact segment endpoints as `OnEdge`
+// instead of `OnVertex`, in both the collinear and the non-parallel branches.
+
+use parry2d::math::Vector;
+use parry2d::shape::SegmentPointLocation;
+use parry2d::utils::{segments_intersection2d, SegmentsIntersection};
+
+#[test]
+fn identical_segments_intersect_on_vertices() {
+    // Exact case from the issue.
+    let a = Vector::new(10.0, 0.0);
+    let b = Vector::new(10.0, 10.0);
+
+    let intersection = segments_intersection2d(a, b, a, b, 0.0).unwrap();
+
+    let SegmentsIntersection::Segment {
+        first_loc1,
+        first_loc2,
+        second_loc1,
+        second_loc2,
+    } = intersection
+    else {
+        panic!("the intersection should be a Segment intersection");
+    };
+
+    assert_eq!(first_loc1, SegmentPointLocation::OnVertex(0));
+    assert_eq!(first_loc2, SegmentPointLocation::OnVertex(0));
+    assert_eq!(second_loc1, SegmentPointLocation::OnVertex(1));
+    assert_eq!(second_loc2, SegmentPointLocation::OnVertex(1));
+}
+
+#[test]
+fn identical_horizontal_segments_intersect_on_vertices() {
+    // Same check through the non-vertical code path of `between()`.
+    let a = Vector::new(-3.0, 2.0);
+    let b = Vector::new(5.0, 2.0);
+
+    let intersection = segments_intersection2d(a, b, a, b, 0.0).unwrap();
+
+    let SegmentsIntersection::Segment {
+        first_loc1,
+        first_loc2,
+        second_loc1,
+        second_loc2,
+    } = intersection
+    else {
+        panic!("the intersection should be a Segment intersection");
+    };
+
+    assert_eq!(first_loc1, SegmentPointLocation::OnVertex(0));
+    assert_eq!(first_loc2, SegmentPointLocation::OnVertex(0));
+    assert_eq!(second_loc1, SegmentPointLocation::OnVertex(1));
+    assert_eq!(second_loc2, SegmentPointLocation::OnVertex(1));
+}
+
+#[test]
+fn crossing_at_endpoint_is_on_vertex() {
+    // Two non-parallel segments touching at seg1's vertex 1 == seg2's vertex 0.
+    // Exercises the `s == 1.0` fix (previously `s == denom`, so `OnVertex(1)`
+    // was unreachable).
+    let a = Vector::new(0.0, 0.0);
+    let b = Vector::new(1.0, 1.0);
+    let c = Vector::new(1.0, 1.0);
+    let d = Vector::new(2.0, 0.0);
+
+    let intersection = segments_intersection2d(a, b, c, d, 0.0).unwrap();
+
+    let SegmentsIntersection::Point { loc1, loc2 } = intersection else {
+        panic!("the intersection should be a Point intersection");
+    };
+
+    assert_eq!(loc1, SegmentPointLocation::OnVertex(1));
+    assert_eq!(loc2, SegmentPointLocation::OnVertex(0));
+}
+
+#[test]
+fn crossing_at_interior_point_is_on_edge() {
+    // Proper crossing in both interiors: still `OnEdge`.
+    let a = Vector::new(-1.0, 0.0);
+    let b = Vector::new(1.0, 0.0);
+    let c = Vector::new(0.0, -1.0);
+    let d = Vector::new(0.0, 1.0);
+
+    let intersection = segments_intersection2d(a, b, c, d, 0.0).unwrap();
+
+    let SegmentsIntersection::Point { loc1, loc2 } = intersection else {
+        panic!("the intersection should be a Point intersection");
+    };
+
+    assert_eq!(loc1, SegmentPointLocation::OnEdge([0.5, 0.5]));
+    assert_eq!(loc2, SegmentPointLocation::OnEdge([0.5, 0.5]));
+}
--- a/crates/parry2d/tests/issue_111_triangle_area_degenerate.rs
+++ b/crates/parry2d/tests/issue_111_triangle_area_degenerate.rs
@@ -1,0 +1,81 @@
+// Regression test for https://github.com/dimforge/parry/issues/111 (2D version).
+//
+// `Triangle::area` now uses half the perp-product magnitude in 2D, which is
+// exact (0.0) when the vertices are bitwise collinear.
+
+use parry2d::math::{Real, Vector};
+use parry2d::shape::Triangle;
+
+/// The previous implementation (Kahan's formula on side lengths).
+fn kahan_area(tri: &Triangle) -> Real {
+    let mut s = [
+        tri.b.distance(tri.a),
+        tri.c.distance(tri.b),
+        tri.a.distance(tri.c),
+    ];
+    s.sort_by(|x, y| x.partial_cmp(y).unwrap());
+    let (c, b, a) = (s[0], s[1], s[2]); // a >= b >= c
+
+    let sqr = (a + (b + c)) * (c - (a - b)) * (c + (a - b)) * (a + (b - c));
+    sqr.max(0.0).sqrt() * 0.25
+}
+
+#[test]
+fn degenerate_triangle_area_is_exactly_zero() {
+    // The issue's collinear triangle, projected to 2D (all points share x, so
+    // use the (y, z) coordinates).
+    let tri = Triangle::new(
+        Vector::new(-2.871, 17.464),
+        Vector::new(1.629, 17.464),
+        Vector::new(-1.521, 17.464),
+    );
+    assert_eq!(tri.area(), 0.0);
+
+    // Two identical vertices.
+    let tri = Triangle::new(
+        Vector::new(2.277, -7.9),
+        Vector::new(-0.57, -8.1),
+        Vector::new(-0.57, -8.1),
+    );
+    assert_eq!(tri.area(), 0.0);
+}
+
+#[test]
+fn area_matches_kahan_for_well_conditioned_triangles() {
+    let tris = [
+        Triangle::new(
+            Vector::new(0.0, 0.0),
+            Vector::new(1.0, 0.0),
+            Vector::new(0.0, 1.0),
+        ),
+        Triangle::new(
+            Vector::new(1.0, 2.0),
+            Vector::new(4.0, 0.0),
+            Vector::new(2.0, 5.0),
+        ),
+        Triangle::new(
+            Vector::new(-3.0, 1.0),
+            Vector::new(0.0, 4.0),
+            Vector::new(5.0, -1.0),
+        ),
+    ];
+
+    for tri in &tris {
+        let expected = kahan_area(tri);
+        let area = tri.area();
+        assert!(
+            (area - expected).abs() <= expected * 1.0e-6,
+            "area {area} != kahan area {expected}"
+        );
+    }
+
+    assert_eq!(
+        Triangle::new(
+            Vector::new(0.0, 0.0),
+            Vector::new(1.0, 0.0),
+            Vector::new(0.0, 1.0),
+        )
+        .area(),
+        0.5
+    );
+}
--- a/crates/parry2d/tests/issue_168_ball_polyline_endpoint_normal.rs
+++ b/crates/parry2d/tests/issue_168_ball_polyline_endpoint_normal.rs
@@ -1,0 +1,80 @@
+// Regression test for https://github.com/dimforge/rapier/issues/168
+//
+// A ball contacting a polyline exactly at (or beyond) its endpoint vertex used
+// to produce a zero contact normal. The ball-vs-convex manifold computation now
+// has a degenerate fallback so the normal is always finite and non-zero.
+
+use parry2d::math::{Pose, Real, Vector};
+use parry2d::query::{ContactManifold, DefaultQueryDispatcher, PersistentQueryDispatcher};
+use parry2d::shape::{Ball, Polyline};
+
+fn check_manifold_normals(pos12: Pose, prediction: Real) -> usize {
+    let polyline = Polyline::new(
+        vec![Vector::new(-10.0, 0.0), Vector::new(10.0, 0.0)],
+        None,
+    );
+    let ball = Ball::new(1.0);
+
+    let dispatcher = DefaultQueryDispatcher;
+    let mut manifolds: Vec<ContactManifold<(), ()>> = Vec::new();
+    let mut workspace = None;
+
+    dispatcher
+        .contact_manifolds(&pos12, &polyline, &ball, prediction, &mut manifolds, &mut workspace)
+        .unwrap();
+
+    let mut num_contacts = 0;
+
+    for manifold in &manifolds {
+        if manifold.points.is_empty() {
+            continue;
+        }
+
+        num_contacts += manifold.points.len();
+
+        for local_n in [manifold.local_n1, manifold.local_n2] {
+            assert!(
+                local_n.is_finite(),
+                "non-finite contact normal: {local_n:?} (pos12: {pos12:?})"
+            );
+            assert!(
+                local_n.length() > 0.9,
+                "degenerate contact normal: {local_n:?} (pos12: {pos12:?})"
+            );
+        }
+    }
+
+    num_contacts
+}
+
+#[test]
+fn ball_beyond_polyline_endpoint() {
+    // The geometry from the issue: the polyline ends at x = 10.0, the ball
+    // (radius 1) is at x = 11.0, y = 1.0, so it barely misses the endpoint but
+    // is within the contact prediction distance.
+    let num_contacts = check_manifold_normals(Pose::translation(11.0, 1.0), 1.0);
+    assert!(num_contacts > 0);
+}
+
+#[test]
+fn ball_aligned_with_polyline_beyond_endpoint() {
+    // The `______o` case from the issue: ball right beside the line, aligned
+    // with it but not touching it.
+    let num_contacts = check_manifold_normals(Pose::translation(11.5, 0.0), 1.0);
+    assert!(num_contacts > 0);
+}
+
+#[test]
+fn ball_centered_on_polyline_endpoint() {
+    // Fully degenerate case: the ball center lies exactly on the endpoint
+    // vertex, so the projection distance is zero.
+    let num_contacts = check_manifold_normals(Pose::translation(10.0, 0.0), 0.1);
+    assert!(num_contacts > 0);
+}
+
+#[test]
+fn ball_overlapping_polyline_endpoint() {
+    // Overlap case without prediction: the ball overlaps the endpoint vertex.
+    let num_contacts = check_manifold_normals(Pose::translation(10.5, 0.5), 0.0);
+    assert!(num_contacts > 0);
+}
--- a/crates/parry2d/tests/issue_168_ball_polyline_endpoint_normal.rs
+++ b/crates/parry2d/tests/issue_168_ball_polyline_endpoint_normal.rs
@@ -9,10 +9,7 @@ use parry2d::query::{ContactManifold, DefaultQueryDispatcher, PersistentQueryDis
 use parry2d::shape::{Ball, Polyline};
 
 fn check_manifold_normals(pos12: Pose, prediction: Real) -> usize {
-    let polyline = Polyline::new(
-        vec![Vector::new(-10.0, 0.0), Vector::new(10.0, 0.0)],
-        None,
-    );
+    let polyline = Polyline::new(vec![Vector::new(-10.0, 0.0), Vector::new(10.0, 0.0)], None);
     let ball = Ball::new(1.0);
 
     let dispatcher = DefaultQueryDispatcher;
@@ -20,7 +17,14 @@ fn check_manifold_normals(pos12: Pose, prediction: Real) -> usize {
     let mut workspace = None;
 
     dispatcher
-        .contact_manifolds(&pos12, &polyline, &ball, prediction, &mut manifolds, &mut workspace)
+        .contact_manifolds(
+            &pos12,
+            &polyline,
+            &ball,
+            prediction,
+            &mut manifolds,
+            &mut workspace,
+        )
         .unwrap();
 
     let mut num_contacts = 0;
--- a/crates/parry2d/tests/issue_180_shape_cast_grazing_none.rs
+++ b/crates/parry2d/tests/issue_180_shape_cast_grazing_none.rs
@@ -11,14 +11,8 @@ use parry2d::query::{self, ShapeCastOptions};
 use parry2d::shape::{Capsule, Cuboid, Segment};
 
 fn cast(pos12: Pose, vel12: Vector, segment: &Segment, cuboid: &Cuboid) -> bool {
-    cast_shapes_support_map_support_map(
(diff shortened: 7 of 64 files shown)


Ughuuu:
Letsgo! so many fixes!

PR #442, opened 2026-08-25
compound pseudo normals (2d and 3d)

Ughuuu:
- Add `CompoundFlags::FIX_INTERNAL_EDGES` to fix normals for compound shape. In godot, we receive a polygon already triangulated with shapes sliding across others. So intersecting points provide normals we can use as pseudonormals.


sebcrozet:
Thank you! The changes on point-projection and ray-casting are unrelated to internal edges fixing, so I removed them. They are still worth it (but perhaps written differently), but will be a breaking change, so I’ll defer to another PR.

PR #446, opened 2026-09-07
Add an analytic ray-capsule intersection test

joonazan:
I need ray-capsule intersections and was surprised that they were really slow because they are implemented via GJK, so I wrote an analytical version.

I did not solve the equations myself. I adapted [Inigo Quilez's GLSL](https://iquilezles.org/articles/intersectors/). The result is more optimized than anything else I can find.

On my machine the analytical version takes about 5 ns per query hit or miss. GJK takes 50 ns on a miss and 400 - 700 ns on a hit.

The analytical version's error is smaller, too. Could be better, though. In extreme corner cases errors of 1e-3 can be observed at scales of a few units using f32.

### DIFF
--- a/src/query/ray/mod.rs
+++ b/src/query/ray/mod.rs
@@ -16,6 +16,7 @@ pub mod ray;
 mod ray_aabb;
 mod ray_ball;
 mod ray_bounding_sphere;
+mod ray_capsule;
 #[cfg(feature = "alloc")]
 mod ray_composite_shape;
 mod ray_cuboid;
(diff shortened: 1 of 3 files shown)


joonazan:
Replaced the initial AI version of the collision check with a handwritten one that is shorter and is much more numerically stable due to cross products instead of dot products followed by `a * b - c * d`. Worst case error didn't improve much but it is way better on average.

sebcrozet:
Thank you!

